PR #102, opened 2026-04-09
fix(security): address post-audit findings

dev-jodee:
## Summary
Roll up the five post-audit security fixes into a single PR so they can land with normal PR CI while preserving finding-by-finding traceability.

## Context
The fixes were originally prepared through the security advisory temporary-fork workflow, but that path does not run normal PR checks. This PR publishes the same aggregate branch against `main` so the repository's standard CI can validate the full set before merge.

## Changes
- harden Crossmint wallet locator URL construction to prevent path retargeting and traversal-style locator inputs
- bind Crossmint signing responses to the requested transaction/message bytes before accepting returned signatures
- harden GCP KMS key resource URL construction and validation
- reject Fireblocks `PROGRAM_CALL` `tx_hash` values as reusable signature artifacts
- make `PrivySigner::pubkey()` fail fast before initialization instead of aliasing to `Pubkey::default()`

### Key Implementation Details
- Crossmint URL construction and response validation changes live in `rust/src/crossmint/mod.rs` and corresponding TypeScript signer/test updates.
- Fireblocks now fails closed when a `PROGRAM_CALL` flow returns only a transaction hash where local signature semantics are required.
- Privy retains its existing initialized behavior but now explicitly rejects uninitialized identity reads.
- The aggregate branch preserves one logical commit per finding for audit traceability.

## Use Cases
- prevent attacker-controlled wallet locator segments from retargeting Crossmint API requests
- prevent accepting mismatched upstream signatures or serialized transactions for different bytes
- prevent transaction broadcast identifiers from being reused as signer-bound Fireblocks signatures
- prevent uninitialized Privy signers from collapsing onto the zero-address identity
- ensure GCP KMS signer resource URLs are built from canonical validated path segments

### TypeScript
- `pnpm -r --filter @solana/keychain-crossmint --filter @solana/keychain-fireblocks --filter @solana/keychain-gcp-kms test:unit`

### Rust
- `cargo test --no-default-features --features all,sdk-v2,unsafe-debug test_privy_`
- `cargo test --no-default-features --features all,sdk-v2,unsafe-debug test_sign_transaction_program_call_rejects_tx_hash_as_signature`
- `cargo test --no-default-features --features all,sdk-v2,unsafe-debug test_program_call_txhash_can_be_foreign_slot0_signature_but_is_rejected`
- `cargo test --no-default-features --features all,sdk-v2,unsafe-debug test_build_wallets_api_url_encodes_raw_slashes_in_wallet_locator`
- `cargo test --no-default-features --features all,sdk-v2,unsafe-debug test_build_wallets_api_url_prevents_dot_segment_retargeting`
- `cargo test --no-default-features --features all,sdk-v2,unsafe-debug test_build_wallets_api_url_double_encodes_encoded_traversal_sequences`

### DIFF
--- a/typescript/packages/crossmint/src/crossmint-signer.ts
+++ b/typescript/packages/crossmint/src/crossmint-signer.ts
@@ -39,7 +39,6 @@ const API_VERSION = '2025-06-09';
 const DEFAULT_API_BASE_URL = 'https://www.crossmint.com/api';
 const DEFAULT_POLL_INTERVAL_MS = 1000;
 const DEFAULT_MAX_POLL_ATTEMPTS = 60;
-const TEST_SIGNER_DERIVED_PUBKEY_ENV = 'TEST_CROSSMINT_SIGNER_DERIVED_PUBKEY';
 
 let base16Encoder: ReturnType<typeof getBase16Encoder> | undefined;
 let base58Decoder: ReturnType<typeof getBase58Decoder> | undefined;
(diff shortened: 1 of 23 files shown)


typescript/packages/crossmint/src/crossmint-signer.ts
```diff
@@ -363,47 +359,65 @@ class CrossmintSigner<TAddress extends string = string> implements SolanaSigner<
...
         if (fromTxId) {
-            return fromTxId;
-        }
-
-        const submitted = response.approvals?.submitted ?? [];
-        for (const approval of submitted) {
-            const signature = decodeSignatureString(approval.signature);
-            if (signature) {
-                return signature;
+            if (!transactionFailed) {
```
amilz: I don't think we want this conditional--I think we should be able to `assertSignatureValid` regardless. If it's bad, it should throw.  as-is, It turns “couldn’t validate onChain.transaction” into “skip validation and trust txId,”

amilz (approved):

PR #110, opened 2026-04-29
Add Openfort backend wallet signer (Rust + TypeScript)

n00m4d:
Add an Openfort backend wallet signer to both the Rust crate and the TypeScript SDK. Openfort holds the Solana private key inside a TEE; this     
  signer authenticates with Openfort's API and delegates signing to `POST /v2/accounts/backend/{accountId}/sign`.


n00m4d:
Close this PR to open a new one from Openfort organization.

PR #132, opened 2026-05-19
fix(rust): roll up Dependabot dependency updates

dev-jodee:
## Summary
- Rolls up the Dependabot PRs that do not require crypto serialization/hash compatibility churn: #125 and #127.
- Moves litesvm and the direct solana-sdk dependency to the 3.x-compatible graph and removes the duplicate litesvm alias.
- Keeps direct bincode pinned at 1.3.3 to match the Solana SDK 3.0 transaction serialization stack.
- Keeps direct hkdf/sha2 on the current compatible stack instead of adding a second direct sha2 alias for RSA-backed signing.

## Not Included
- Does not include #123. The direct bincode dependency remains 1.3.3 because Solana SDK 3.0.0 and related Solana crates still use bincode 1.3.3 for transaction serialization.
- Does not include #124 or #126. Updating hkdf to 0.13 requires sha2 0.11 for Crossmint HKDF usage, but rsa 0.9 still uses the sha2/digest 0.10 trait stack for DFNS RSA signing. Keeping direct sha2 on 0.10.9 avoids a duplicate direct sha2 dependency in this crate.

## Test Plan
- cargo check --locked --features all,sdk-v2
- cargo check --locked --no-default-features --features all,sdk-v3
- cargo clippy --locked --all-targets --features all,sdk-v2,unsafe-debug,integration-tests -- -D warnings
- cargo clippy --locked --all-targets --no-default-features --features all,sdk-v3,unsafe-debug,integration-tests -- -D warnings
- cargo test --locked --features all,sdk-v2,unsafe-debug test_sign_challenge_rsa
- cargo fmt --all -- --check
- git diff --check

## Breaking Changes
- The sdk-v2 feature now resolves the direct solana-sdk dependency to 3.0.0 as requested by Dependabot PR #127; consumers depending on Solana SDK 2.x through that feature should review before merging.

Supersedes #125 and #127.


dev-jodee:
Closing this rollup because preserving the public sdk-v2/sdk-v3 split leaves no useful Dependabot rollup here.

Findings from the rollup attempt:
- #123 bincode should not move past 1.3.3 because the Solana SDK 3.0 transaction serialization stack still uses bincode 1.3.3.
- #124/#126 hkdf + sha2 would require either a duplicate direct sha2 dependency or a broader RSA/digest migration; keeping direct sha2 on 0.10.9 avoids that.
- #127 solana-sdk cannot be rolled into the sdk-v2 feature because sdk-v2 must continue resolving Solana SDK 2.x.
- #125 litesvm is tied to that same SDK graph: sdk-v2 needs the 0.7 line, while sdk-v3 uses the 0.8 alias.

Net: these need to be handled one by one, and some should be intentionally closed/ignored rather than rolled up.

PR #145, opened 2026-05-29
Add Utila signer

GuiBibeau:
## Summary
- add Utila as a Rust and TypeScript signer backend
- wire Utila into the keychain factory, exports, docs, and CI package lists
- support Solana transaction signing via Utila serialized transactions; keep message signing explicitly unsupported

## Validation
- pnpm -F @solana/keychain-utila test
- pnpm -F @solana/keychain-utila test:integration
- cargo test --no-default-features --features utila,integration-tests,sdk-v2 test_utila_ -- --nocapture --test-threads=1
- live Utila co-signer signed devnet test transactions successfully

### DIFF
--- a/rust/src/utila/mod.rs
+++ b/rust/src/utila/mod.rs
@@ -1,0 +1,947 @@
+//! Utila API signer integration
+
+mod types;
+
+use crate::sdk_adapter::{Pubkey, Signature, Transaction, VersionedTransaction};
+use crate::traits::{SignTransactionResult, SignedTransaction};
+use crate::transaction_util::TransactionUtil;
+use crate::{error::SignerError, http_client_config::HttpClientConfig, traits::SolanaSigner};
+use base64::{engine::general_purpose::STANDARD, Engine};
+use chrono::{Duration, Utc};
+use jsonwebtoken::{encode, Algorithm, EncodingKey, Header};
+use serde::Serialize;
+use std::{str::FromStr, sync::Arc};
+use types::{
+    InitiateTransactionDetails, InitiateTransactionRequest, SolanaSerializedTransaction,
+    TransactionEnvelope, TransactionState, UtilaTransaction, WalletResponse,
+};
+
+const DEFAULT_API_BASE_URL: &str = "https://api.utila.io";
+const UTILA_API_AUDIENCE: &str = "https://api.utila.io/";
+const DEFAULT_POLL_INTERVAL_MS: u64 = 1000;
+const DEFAULT_MAX_POLL_ATTEMPTS: u32 = 60;
+const TOKEN_TTL_MINUTES: i64 = 55;
+const AVAILABILITY_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
+
+/// Configuration for creating a UtilaSigner.
+#[derive(Clone)]
+pub struct UtilaSignerConfig {
+    pub service_account_email: String,
+    pub service_account_private_key_pem: String,
+    pub vault_id: String,
+    pub wallet_id: String,
+    /// Utila network resource name, e.g. `networks/solana-devnet`.
+    pub network: String,
+    pub api_base_url: Option<String>,
+    pub poll_interval_ms: Option<u64>,
+    pub max_poll_attempts: Option<u32>,
+    /// Utila signer resource names. Defaults to `users/{service_account_email}`.
+    pub designated_signers: Option<Vec<String>>,
+    pub http_client_config: Option<HttpClientConfig>,
+}
+
+/// Utila-backed signer using an existing Solana wallet.
+#[derive(Clone)]
+pub struct UtilaSigner {
+    service_account_email: String,
+    signing_key: Arc<EncodingKey>,
+    vault_id: String,
+    wallet_id: String,
+    network: String,
+    api_base_url: String,
+    client: reqwest::Client,
+    public_key: Option<Pubkey>,
+    poll_interval_ms: u64,
+    max_poll_attempts: u32,
+    designated_signers: Vec<String>,
+}
+
+impl std::fmt::Debug for UtilaSigner {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("UtilaSigner")
+            .field("public_key", &self.public_key)
+            .field("vault_id", &self.vault_id)
+            .field("wallet_id", &self.wallet_id)
+            .field("network", &self.network)
+            .finish_non_exhaustive()
+    }
+}
+
+#[derive(Serialize)]
+struct UtilaAccessTokenClaims<'a> {
+    sub: &'a str,
+    aud: &'a str,
+    exp: i64,
+}
+
+impl UtilaSigner {
+    /// Create a new Utila signer.
+    ///
+    /// You must call `init()` after construction to fetch the wallet address.
+    pub fn new(config: UtilaSignerConfig) -> Result<Self, SignerError> {
+        validate_required("service_account_email", &config.service_account_email)?;
+        validate_required(
+            "service_account_private_key_pem",
+            &config.service_account_private_key_pem,
+        )?;
+        validate_required("vault_id", &config.vault_id)?;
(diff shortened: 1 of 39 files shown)


GuiBibeau:
@greptileai review

typescript/packages/dfns/src/auth.ts
amilz, line 1: What's thinking for touching dfns in this PR?

dev-jodee: reusable util for normalizing .pem

rust/src/utila/mod.rs
```diff
@@ -0,0 +1,947 @@
...
+    async fn fetch_wallet(&self) -> Result<WalletResponse, SignerError> {
+        let path = format!("/v2/vaults/{}/wallets/{}", self.vault_id, self.wallet_id);
+        self.get_json(&path, "fetch_wallet").await
+    }
+
+    async fn initiate_transaction(
+        &self,
+        raw_transaction: String,
+    ) -> Result<UtilaTransaction, SignerError> {
+        let path = format!("/v2/vaults/{}{}", self.vault_id, "/transactions:initiate");
```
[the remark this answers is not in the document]
dev-jodee: Already handled — both `fetch_wallet` and `initiate_transaction` use `encode_uri_component` for `vault_id`/`wallet_id`, same as `get_transaction`. Stale against current HEAD.

rust/src/utila/mod.rs
```diff
@@ -0,0 +1,1007 @@
...
+        matches!(result, Ok(Ok(_)))
+    }
+}
+
+#[async_trait::async_trait]
+impl SolanaSigner for UtilaSigner {
+    fn pubkey(&self) -> Pubkey {
+        self.public_key.expect("UtilaSigner not initialized")
+    }
+
```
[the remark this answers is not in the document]
dev-jodee: Consistent with the existing convention — `OpenfortSigner` and `FireblocksSigner` use the same `self.public_key.expect("... not initialized")` in `pubkey()`. The atomic safe path already exists via `Signer::from_utila(config).await`, which runs `new()` + `init()` together. Keeping parity with the sibling backends rather than diverging in this PR.

typescript/packages/utila/src/utila-signer.ts
```diff
@@ -0,0 +1,516 @@
...
+        const initiated = await this.initiateTransaction(rawTransaction);
+        const signed = await this.pollSignedTransaction(initiated);
+        const rawSignedTransaction = signed.solanaTransaction?.rawTransaction;
+        if (!rawSignedTransaction) {
+            throwSignerError(SignerErrorCode.SIGNING_FAILED, {
+                message: 'Utila signed transaction response missing solanaTransaction.rawTransaction',
+            });
+        }
+
+        return await this.extractSignatureFromRawTransaction(rawSignedTransaction, transaction.messageBytes);
```
amilz: can leverage `extractSignatureFromWireTransaction` from core package and just return a `SignatureDictionary`

typescript/packages/utila/src/utila-signer.ts
```diff
@@ -0,0 +1,516 @@
...
+
+export async function createUtilaAccessToken(
+    serviceAccountEmail: string,
+    privateKey: ImportedPrivateKey,
+): Promise<string> {
+    try {
+        return await new SignJWT({})
+            .setProtectedHeader({ alg: 'RS256' })
+            .setSubject(serviceAccountEmail)
+            .setAudience(UTILA_API_AUDIENCE)
```
amilz, line 70: Not familiar w/ utiliia's model, but just confirming we wouldn't need to ever pass a custom `apiBaseUrl` in here if one was supplied on signer initiation?

typescript/packages/utila/src/utila-signer.ts
```diff
@@ -0,0 +1,516 @@
...
+        const walletId = trimWalletId(config.walletId);
+        const designatedSigners = config.designatedSigners ?? [`users/${config.serviceAccountEmail}`];
+        const wallet = await fetchWallet({
+            apiBaseUrl,
+            privateKey,
+            serviceAccountEmail: config.serviceAccountEmail,
+            vaultId,
+            walletId,
+        });
+        const address = wallet.wallet?.solanaDetails?.address;
```
amilz: treat bad JSON payloads as remote API errors, not raw TypeErrors.

```suggestion
        const address = wallet?.wallet?.solanaDetails?.address;
```

typescript/packages/utila/src/utila-signer.ts
```diff
@@ -0,0 +1,516 @@
...
+    } catch (error) {
+        throwSignerError(SignerErrorCode.PARSING_ERROR, {
+            cause: error,
+            message: 'Failed to parse Utila wallet response',
+        });
+    }
+}
+
+function parseTransactionEnvelope(payload: UtilaTransactionEnvelope, context: string): UtilaTransaction {
+    const transaction = payload.transaction;
```
amilz: in case bad json response

```suggestion
    const transaction = payload?.transaction;
```

amilz (commented):
ts lgtm. couple nits thanks @GuiBibeau 
i need to update our skill. if you don't mind adding the new signer to typescript/scripts/test-treeshake-umbrella.mjs

typescript/packages/utila/src/utila-signer.ts
dev-jodee: Done in 07b1d3f — `extractSignatureFromRawTransaction` now returns a `SignatureDictionary` via core `createSignatureDictionary`. Kept the message-bytes equality check + `assertSignatureValid` rather than swapping wholesale to `extractSignatureFromWireTransaction`, since that core helper skips both — Utila intentionally verifies the returned message matches and the signature is valid (parity with the Rust signer's `signature.verify`).

typescript/packages/utila/src/utila-signer.ts
dev-jodee: Confirmed via Utila docs (https://docs.utila.io/reference/authentication): the JWT `aud` is always `https://api.utila.io/` — a fixed audience identifier, independent of the HTTP endpoint. So a custom `apiBaseUrl` never needs to flow into the token; `aud` stays constant. Both Rust and TS already match the documented model. No change needed.

typescript/packages/utila/src/utila-signer.ts
dev-jodee: Done in 07b1d3f — added `wallet?.` so an unexpected/null JSON payload falls through to the existing `INVALID_PUBLIC_KEY` throw instead of a raw TypeError. (Malformed JSON was already caught upstream as `PARSING_ERROR`; this covers valid-but-unexpected shapes.)

typescript/packages/utila/src/utila-signer.ts
dev-jodee: Done in 07b1d3f — added `payload?.transaction` so a null/unexpected payload throws `REMOTE_API_ERROR` via the existing name/state guard instead of a raw TypeError.

dev-jodee (dismissed):

typescript/packages/utila/src/utila-signer.ts
amilz: Thx--main point was that we could leverage the util from core instead of a local method. WDYT?

https://github.com/solana-foundation/solana-keychain/blob/9b3a7070863f6954766a65cd97d1aacb5b9ec42c/typescript/packages/core/src/utils.ts#L84

dev-jodee: Good call. Done in ab733ee. Now use core `extractSignatureFromWireTransaction` + `assertSignatureValid` against `transaction.messageBytes` — same pattern as privy/cdp. Dropped local method + explicit message-bytes compare. Compare redundant: `assertSignatureValid` against our local message already reject tampered/different signed message crypto-side.

dev-jodee:
Utila needs co signer to be running, so integration tests wont run on the CI (just like turnkey doesn't because it requires the vault running). I've tested it locally with the full integration tests flow and it worked.

PR #152, opened 2026-06-09
fix(crossmint): sequential signing + scoped approval for managed flow

amilz:
## Summary

Two correctness fixes for the Crossmint signer's managed transaction flow, plus a doc-only note documenting an intentional verification decision.

- **Sign transactions sequentially** (`signTransactions`): replace `Promise.all` with a sequential loop. Crossmint auto-broadcasts each created transaction server-side once approved, so concurrent batch submission risks landing some transactions while abandoning siblings on a mid-batch failure — leading to duplicate spends on retry. Sequential execution stops on the first error before any further transaction is created.
- **Approve only this signer's pending challenge, once**: on multi-approver wallets `approvals.pending[]` can contain challenges for other approvers. Match the pending entry by this signer's locator instead of blindly signing `pending[0]`, and guard with an `approvalSubmitted` flag so the poll loop can't re-submit the same approval.
- **Docs**: class-level note + guard comment clarifying that the Crossmint signer is *broadcast-managed* (not a self-broadcast partial signer), so it intentionally verifies returned signatures against the bytes Crossmint actually signed rather than the caller's original `messageBytes`.

## Why the verification stays lenient

An earlier draft of this work also flipped signature verification to check against the caller's original transaction bytes. We dropped that change: per [Crossmint's docs](https://docs.crossmint.com/api-reference/wallets/create-transaction), Crossmint broadcasts server-side and rewrites the transaction before signing (gas sponsorship sets `feePayer`, priority-fee optimization, its own recent blockhash). A strict check against the caller's bytes would false-reject every legitimately landed transaction — and would hard-fail the default gas-sponsored flow (`SIGNING_FAILED`). The lenient check (introduced in #102) is the docs-correct behavior: it still proves Crossmint signed a coherent transaction with our key without rejecting legitimate rewrites. The new comments record this so it isn't re-flagged.

## Test plan

- `pnpm --filter @solana/keychain-crossmint typecheck` — passes
- `pnpm --filter @solana/keychain-crossmint test` — 31 passed, 3 skipped
- `just ts-fmt` — clean, no changes

closes TOO-482

### DIFF
--- a/typescript/packages/crossmint/src/__tests__/crossmint-signer.test.ts
+++ b/typescript/packages/crossmint/src/__tests__/crossmint-signer.test.ts
@@ -673,8 +673,8 @@ describe('CrossmintSigner', () => {
                             approvals: {
                                 // pending[0] belongs to ANOTHER approver.
                                 pending: [
-                                    { signer: OTHER_SIGNER, message: OTHER_MESSAGE_B58 },
-                                    { signer: OUR_SIGNER, message: OUR_MESSAGE_B58 },
+                                    { signer: { locator: OTHER_SIGNER }, message: OTHER_MESSAGE_B58 },
+                                    { signer: { locator: OUR_SIGNER }, message: OUR_MESSAGE_B58 },
                                 ],
                             },
                         }),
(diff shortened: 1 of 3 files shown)


dev-jodee (approved):
=

PR #153, opened 2026-06-10
fix(fireblocks): reject useProgramCall at construction, before any broadcast

amilz:
## Summary

Fireblocks `PROGRAM_CALL` signing is incompatible with the `SolanaSigner` contract: it broadcasts the transaction on-chain and only returns a broadcast transaction id, never a reusable signer-bound signature over the local message bytes. The old code attempted it and then always threw `SIGNING_FAILED` — *after* the on-chain broadcast — risking duplicate spends on retry.

This PR rejects `useProgramCall: true` at construction with a `CONFIG_ERROR` (before any network call) and removes the dead PROGRAM_CALL path. RAW signing (sign bytes, caller broadcasts) is the only supported mode and the default.

https://developers.fireblocks.com/reference/interact-with-solana-programs

## Changes

- Reject `useProgramCall: true` in the constructor, fail-closed before any I/O.
- Remove the PROGRAM_CALL signing path, `ProgramCallExtraParameters`, and `txHash` handling.
- Mark `useProgramCall` `@deprecated` — retained for now so existing callers get a clear error instead of silent behavior change; removal planned in a future major.
- Update tests (construction-time rejection, no network call) and README.

## Notes

- RAW already provides no-broadcast, exact-message-bytes signing with a documented response — `signOnly`/`useDurableNonce` on PROGRAM_CALL would only reconstruct RAW the hard way.
- Rust parity: `rust/src/fireblocks/` still has the broadcast-then-fail path; if we merge this, we should do same for rust side @dev-jodee 

Closes TOO-483

### DIFF
--- a/rust/src/fireblocks/mod.rs
+++ b/rust/src/fireblocks/mod.rs
@@ -10,12 +10,10 @@ use crate::{
     error::SignerError, http_client_config::HttpClientConfig, traits::SolanaSigner,
     transaction_util::TransactionUtil,
 };
-use base64::{engine::general_purpose::STANDARD, Engine};
 use std::{str::FromStr, sync::Arc};
 use types::{
-    CreateTransactionRequest, CreateTransactionResponse, ExtraParameters,
-    ProgramCallExtraParameters, RawExtraParameters, RawMessage, RawMessageData,
-    TransactionResponse, TransactionSource, VaultAddressesResponse,
+    CreateTransactionRequest, CreateTransactionResponse, RawExtraParameters, RawMessage,
+    RawMessageData, TransactionResponse, TransactionSource, VaultAddressesResponse,
 };
 
 use crate::signature_util::EXPECTED_SIGNATURE_LENGTH;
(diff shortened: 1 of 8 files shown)


dev-jodee (commented):
code itself looks fine, but now fireblocks and typescript don't act the same way for fireblocks, we should keep parity. And this is a breaking change right, did we confirm 100% program call wont work? If people are currently using it and we block it and it breaks their signer that could be annoying

amilz:
> code itself looks fine, but now fireblocks and typescript don't act the same way for fireblocks, we should keep parity. And this is a breaking change right, did we confirm 100% program call wont work? If people are currently using it and we block it and it breaks their signer that could be annoying

rn we sign and send...then always throw. so the result of this PR is the same from our API surface...user gets an error. difference here is that we're preventing that transaction from getting sent by fireblocks...which i think makes sense. 
agree if we do this here, we should do it for rust as well. 
https://github.com/solana-foundation/solana-keychain/blob/8cb06038ae6831702a2f3cfb2a1746b7e3ba2163/typescript/packages/fireblocks/src/fireblocks-signer.ts#L377-L386

dev-jodee (approved):

PR #162, opened 2026-06-12
feat(keychain-ts): add @solana/keychain-kit-plugin package (TOO-496)

amilz:
Adds @solana/keychain-kit-plugin, a Kit client plugin package that creates a keychain signer from any backend config and installs it on the client.

Each plugin accepts the same backend-tagged configuration as `createKeychainSigner` from `@solana/keychain`:

```ts
import { createClient } from '@solana/kit';
import { keychainSigner } from '@solana/keychain-kit-plugin';

const client = await createClient().use(
    keychainSigner({
        backend: 'privy',
        appId: process.env.PRIVY_APP_ID!,
        appSecret: process.env.PRIVY_APP_SECRET!,
        walletId: process.env.PRIVY_WALLET_ID!,
    }),
);

client.payer; // SolanaSigner — also a Kit TransactionSigner
client.identity; // same signer instance
```

### Plugin variants

| Plugin                    | Sets                                          |
| ------------------------- | --------------------------------------------- |
| `keychainSigner(config)`  | Both `payer` **and** `identity` (same signer) |
| `keychainPayer(config)`   | Only `payer`                                  |
| `keychainIdentity(config)` | Only `identity`  

- Backend packages remain dynamically imported through the umbrella, so bundles only include the backend a config dispatches to; the package is agadoo-verified tree-shakable.
- Wires the package into CI (build, typecheck, lint, format, unit tests, treeshake — gated on all signers enabled, like the umbrella), the publish workflow, the `release-ts` version-bump list, and the release skill (whose package list was also missing `memory`/`openfort`/`utila`).
- Peer dependency: `@solana/kit >=6.8.0` (`extendClient`/`createClient` landed in 6.8). 

Linear: TOO-496

@lorisleiva would appreciate a quick look if you have a minute. Also WDYT about naming `@solana/kit-plugin-keychain` or would you rather reserve that for the kit-plugins repo?

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -99,7 +99,7 @@ git commit -m "chore: bump rust version to vX.Y.Z"
 
 ```bash
 cd typescript
-for pkg in core aws-kms cdp dfns fireblocks gcp-kms para privy turnkey vault keychain test-utils crossmint; do
+for pkg in core aws-kms cdp dfns fireblocks gcp-kms memory openfort para privy turnkey utila vault keychain kit-plugin test-utils crossmint; do
   cd packages/${pkg} && npm version "A.B.C" --no-git-tag-version && cd ../..
 done
 npm version "A.B.C" --no-git-tag-version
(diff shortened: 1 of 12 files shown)


dev-jodee (approved):

lorisleiva (approved):
Sorry for the late review. Looks great to me!

PR #175, opened 2026-07-02
feat(ledger): add Ledger hardware-wallet signer backend

ledgicr:
> **Draft** — the backend is complete and hardware-validated; opening early so the team can review while two follow-ups land (Gen5/sdk-v4 dependency relaxation, and commit signing). See "Status" at the bottom.

## What

Adds a **`ledger` signer backend** to `solana-keychain` — a hardware-wallet backend alongside the existing software/KMS/custody signers, implementing `SolanaSigner` on top of [`solana-remote-wallet`](https://crates.io/crates/solana-remote-wallet). The private key never leaves the device; every payment is confirmed on-screen.

- `Signer::from_ledger(...)` + `Keystore`/enum wiring, gated behind a new `ledger` feature.
- `SignerError::UserRejected` for on-device rejection.
- Unit tests (derivation, byte bridge, error mapping) + a hardware integration test that self-skips when no device is present.

## Design notes

- **Threading.** `solana-remote-wallet` is single-threaded (`Rc`-based, non-`Sync` hidapi handle), but `SolanaSigner` is async + `Send + Sync`. All device I/O is therefore confined to a dedicated **device-actor thread**; pubkey/signature cross the boundary as raw bytes. This keeps the trait contract intact without making the whole client single-threaded.
- **Signing.** `sign_transaction` routes through the device's transaction-parsing APDU; `sign_message` uses the Solana off-chain-message envelope. (A hardware wallet cannot raw-ed25519-sign arbitrary bytes the way a software key can — it signs a transaction or an off-chain message.)

## Fixes found only by testing on real hardware (a Nano device)

- **Select the device by `Manufacturer::Ledger`, not model** — the model string is `"nano-gen5"`/`"stax"`/etc., never `"ledger"`, so a model-substring match never fires.
- **Default derivation path `m/44'/501'/0'`** to match **Ledger Live** — the 4-component CLI style (`…/0'/0'`) derives a *different* address than the one users fund.
- **On-device rejection → `UserRejected`** — the device returns APDU `0x6985` as a nested `LedgerError::UserCancel`, which is easy to miss and otherwise surfaces as a generic signing failure.
- **`derive(Debug)`** on the unified `Signer` enum.

## Validated on a real Nano device

Connectivity, address derivation, and **transaction signing** (device displays → user approves → valid signature), plus account registration with on-device address confirmation. Driven end-to-end through `pay-kit` and the `pay` CLI; a Ledger signed an MPP charge in sandbox (see solana-foundation/pay-kit#206).

## Scope of this PR (sdk-v3, interim)

Targets the **`sdk-v3`** feature line with `solana-remote-wallet 3.1.14`. This covers **Nano S / X / S+ / Stax / Flex** — every Ledger device **except the Nano Gen5**.

**Gen5 is a deliberate follow-up, not an omission.** Gen5's USB PID (`0x8000`) is only present in `solana-remote-wallet` **4.1.x** (agave #10890) — not on 4.0.x, and not on the EOL 3.1 line (our agave 3.1 backport #13553 was closed "3.1 is eol"). Moving the backend to 4.1 requires the `sdk-v4` line, which is currently blocked by a dependency conflict:

- `solana-remote-wallet 4.1` → `solana-pubkey 4.2`
- `litesvm` (sim tests) → `solana-loader-v4-program 4.0` → `solana-pubkey ~4.1`

These are disjoint, and I verified (2026-07-02) that making litesvm opt-in does **not** clear it — Cargo locks a single `solana-pubkey` for the whole manifest regardless of features. The clean unblock is `litesvm`/`solana-loader-v4-program` shipping on `solana-pubkey 4.2` (the relaxation Aursen has in flight). Once that lands, the Gen5/sdk-v4 bump is a small follow-up (the backend code is ready — WIP branch exists).

## Status / checklist before ready-for-review
- [x] Backend implemented, builds green on `memory,ledger,sdk-v3`, unit tests pass, hardware-validated.
- [ ] **Commit signing** — will sign the commits (this PR's commits are currently unsigned).
- [ ] **Gen5/sdk-v4** — follow-up once `solana-loader-v4-program`/`litesvm` are on `solana-pubkey 4.2`.

Happy to adjust the structure (feature layout, test harness, naming) to fit the repo's conventions — flag anything.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -45,6 +45,9 @@ openfort = [
     "dep:hex",
 ]
 utila = ["dep:reqwest", "dep:jsonwebtoken", "dep:chrono"]
+# Ledger hardware-wallet signing over USB-HID. Requires sdk-v3 (matches the
+# solana-* 3.x crates solana-remote-wallet pins). Not yet supported on sdk-v2/v4.
+ledger = ["dep:solana-remote-wallet", "dep:solana-derivation-path"]
 all = [
     "memory",
     "vault",
@@ -85,6 +88,15 @@ solana-shred-version-v3 = { package = "solana-shred-version", version = "=3.0.0"
 # Pin solana-signature to avoid v3.3.0 where DecodeError doesn't implement std::error::Error
 solana-signature = { version = ">=3.1, <3.5", optional = true }
 solana-sdk-v4 = { package = "solana-sdk", version = "=4.0.1", optional = true }
+
+# Ledger hardware wallet (USB-HID). 3.x line pins solana-pubkey ^3.0.0 /
+# solana-signature ^3.1.0 / solana-derivation-path ^3.0.0 — aligns with sdk-v3.
+# Default features pull a hidapi HID backend (linux-static-hidraw on Linux,
+# IOKit on macOS, native on Windows); agave-unstable-api silences the 3.1
+# deprecation gate (it becomes mandatory in 4.0).
+solana-remote-wallet = { version = "3.1.14", optional = true, features = ["agave-unstable-api"] }
+# DerivationPath type used by the ledger backend (same version solana-remote-wallet pins).
+solana-derivation-path = { version = "3.0.0", optional = true }
 
 # Core dependencies
 async-trait = "0.1.89"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -47,7 +47,11 @@ openfort = [
 utila = ["dep:reqwest", "dep:jsonwebtoken", "dep:chrono"]
 # Ledger hardware-wallet signing over USB-HID. Requires sdk-v3 (matches the
 # solana-* 3.x crates solana-remote-wallet pins). Not yet supported on sdk-v2/v4.
-ledger = ["dep:solana-remote-wallet", "dep:solana-derivation-path"]
+ledger = [
+    "dep:solana-remote-wallet",
+    "dep:solana-derivation-path",
+    "dep:solana-offchain-message",
+]
 # NOTE: `ledger` is intentionally omitted here. It requires `sdk-v3` (enforced by
 # a `compile_error!` in lib.rs), so folding it into `all` would break the common
 # `--features all` build under the default `sdk-v2`. Enable it explicitly with
@@ -101,6 +105,10 @@ solana-sdk-v4 = { package = "solana-sdk", version = "=4.0.1", optional = true }
 solana-remote-wallet = { version = "3.1.14", optional = true, features = ["agave-unstable-api"] }
 # DerivationPath type used by the ledger backend (same version solana-remote-wallet pins).
 solana-derivation-path = { version = "3.0.0", optional = true }
+# Off-chain message envelope for sign_message: a hardware wallet signs the
+# serialized OffchainMessage (\xffsolana offchain …), not raw bytes. Same 3.0.0
+# solana-remote-wallet already pulls, so no extra resolution.
+solana-offchain-message = { version = "3.0.0", optional = true }
 
 # Core dependencies
 async-trait = "0.1.89"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -23,9 +23,10 @@ fireblocks = [
     "dep:uuid",
     "dep:hex",
     "dep:chrono",
+    "dep:tokio",
 ]
-para = ["dep:reqwest", "dep:hex"]
-crossmint = ["dep:reqwest", "dep:hkdf", "dep:sha2", "dep:ed25519-dalek"]
+para = ["dep:reqwest", "dep:hex", "dep:tokio"]
+crossmint = ["dep:reqwest", "dep:hkdf", "dep:sha2", "dep:ed25519-dalek", "dep:tokio"]
 gcp_kms = ["dep:google-cloud-kms-v1", "dep:google-cloud-auth", "dep:reqwest"]
 cdp = [
     "dep:reqwest",
@@ -44,13 +45,17 @@ openfort = [
     "dep:sha2",
     "dep:hex",
 ]
-utila = ["dep:reqwest", "dep:jsonwebtoken", "dep:chrono"]
+utila = ["dep:reqwest", "dep:jsonwebtoken", "dep:chrono", "dep:tokio"]
 # Ledger hardware-wallet signing over USB-HID. Requires sdk-v3 (matches the
 # solana-* 3.x crates solana-remote-wallet pins). Not yet supported on sdk-v2/v4.
+# tokio is optional crate-wide; the actor bridge uses tokio::task::spawn_blocking,
+# so pull in tokio with its `rt` feature explicitly.
 ledger = [
     "dep:solana-remote-wallet",
     "dep:solana-derivation-path",
     "dep:solana-offchain-message",
+    "dep:tokio",
+    "tokio/rt",
 ]
 # NOTE: `ledger` is intentionally omitted here. It requires `sdk-v3` (enforced by
 # a `compile_error!` in lib.rs), so folding it into `all` would break the common
@@ -115,7 +120,10 @@ async-trait = "0.1.89"
 thiserror = "2.0.17"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.145"
-tokio = { version = "1.52.3", features = ["full"] }
+# Only the polling remote backends (fireblocks, para, crossmint, utila) need
+# tokio, and only its timers. Keeping it optional and lean lets runtime-less
+# consumers (wasm32-unknown-unknown, mobile FFI) build the memory signer.
+tokio = { version = "1.52.3", default-features = false, features = ["time"], optional = true }
 log = "0.4.28"
 bs58 = "0.5.1"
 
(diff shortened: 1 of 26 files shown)


ledgicr:
Thanks @greptile — all four addressed in the latest push (`91ced15..711e13b`):

1. **`from_ledger` blocking the runtime (P1):** `Signer::from_ledger` is now `async` and runs the blocking `connect()` on the blocking pool via `spawn_blocking`. `connect()`'s docs now call out that it blocks (until a button press when `confirm_pubkey_on_device` is set) and point async callers to the factory. README example updated to `.await`.
2. **`ledger` absent from `all`:** added a comment at the `all` definition explaining the deliberate omission — `ledger` requires `sdk-v3`, so `all` + the default `sdk-v2` would fail the `compile_error!` gate. Enable it with `--features ledger,sdk-v3`.
3. **Multi-device selection:** `connect`/`from_ledger` now take an optional `host_device_path`. With several Ledgers attached and none specified, we return `NotAvailable` listing each device's host path (and model) so the caller can retry with a specific one — instead of silently picking the first enumerated device.
4. **No-device test robustness:** `initialize_wallet_manager()` failures now map to `NotAvailable` (an availability failure, not a signing failure), and `connect_without_device_*` accepts any `Err` — its contract is "fails cleanly, never hangs/panics," which shouldn't be tied to a specific variant on HID-less CI runners.

All commits signed; `cargo test` / `clippy` / `fmt` green on `memory,ledger,sdk-v3`, and the default (ledger-off) build is unaffected. Nice catches — thanks for the review.

(Context: the `sdk-v3` target is interim; Gen5/`sdk-v4` follows once `solana-loader-v4-program` is on `solana-pubkey 4.2`, per the PR description.)

ledgicr:
Good catches on both off-chain-message comments — they're the same root cause and it was a real bug, fixed in `efdaff9`.

You were right that `sign_offchain_message` sends the bytes straight to the device APDU without building the envelope, so passing the raw payload was wrong on two counts: the device would reject/sign the wrong content, and the test verified against the raw bytes (so it never actually passed on hardware — our prior device validation only exercised *transaction* signing).

Fix:
- `sign_message` now wraps the payload in a Solana `OffchainMessage` (v0) and signs the **serialized** form (the `\xffsolana offchain` envelope). Added `solana-offchain-message 3.0.0` — the same version `solana-remote-wallet` already pulls, so no new resolution.
- The integration test verifies against `OffchainMessage::new(0, msg)?.serialize()?`, not the raw bytes.
- Documented the deviation explicitly, per your suggestion: the `sign_message` doc on both the `SolanaSigner` trait and the Ledger impl now spells out that hardware backends sign the off-chain envelope (verifiable against the serialized message), unlike the software backends' raw-bytes signatures — so verifiers that accept hardware signers must be off-chain-message aware.

This is inherent to hardware signing rather than something we can make byte-identical to the software backends, so it's called out as a documented contract deviation. (Device re-validation of the off-chain path is pending — the transaction path is already hardware-confirmed.)

On the `??` vs `.and_then(|r| r)?` note — agree it's purely stylistic; leaving the idiomatic `??`.

Thanks again — genuinely useful review.

rust/src/tests/test_ledger_integration.rs
```diff
@@ -0,0 +1,84 @@
...
+        assert!(
+            signature.verify(&signer.pubkey().to_bytes(), &serialized),
+            "signature must verify against the serialized off-chain message"
+        );
+    }
+
+    #[tokio::test]
+    #[cfg(feature = "integration-tests")]
+    async fn test_ledger_sign_transaction() {
+        use crate::test_util::create_test_transaction;
```
dev-jodee, line 71: nit: let's keep the `use` at the top of the file instead of in functions

rust/src/lib.rs
```diff
@@ -375,6 +391,35 @@ impl Signer {
...
+        let host_device_path = host_device_path.map(str::to_string);
+        let signer = tokio::task::spawn_blocking(move || {
+            LedgerSigner::connect(
+                derivation_path.as_deref(),
+                confirm_pubkey_on_device,
+                host_device_path.as_deref(),
+            )
+        })
+        .await
+        .map_err(|e| SignerError::Other(format!("Ledger connect task failed: {e}")))??;
```
dev-jodee, line 420: just confirming the "e" won't contain sensitive info?

rust/Cargo.toml
```diff
@@ -45,6 +45,17 @@ openfort = [
     "dep:hex",
 ]
 utila = ["dep:reqwest", "dep:jsonwebtoken", "dep:chrono"]
+# Ledger hardware-wallet signing over USB-HID. Requires sdk-v3 (matches the
+# solana-* 3.x crates solana-remote-wallet pins). Not yet supported on sdk-v2/v4.
+ledger = [
```
dev-jodee, line 53: Need to rebase for the merge conflicts and since pr  #180, tokio is an optional dep, so need to be added here as well :)

rust/src/ledger/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    // here — these unit tests only cover the pure logic that needs no device.
+
+    #[test]
+    fn default_derivation_path_is_solana_bip44() {
+        let path = DerivationPath::from_absolute_path_str(DEFAULT_DERIVATION_PATH);
+        assert!(path.is_ok(), "default derivation path must parse");
+    }
+
+    #[test]
+    fn signature_bytes_roundtrips() {
```
dev-jodee, line 395: for the unit test you'll need to add ledger in the CI actions

rust/src/ledger/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// payload is wrapped in the Solana off-chain envelope (the
+    /// `\xffsolana offchain` signing domain + header) and the device signs that.
+    /// The returned signature therefore covers the **serialized
+    /// `OffchainMessage`**, not the raw `message` bytes. Verify it against the
+    /// serialized form (`OffchainMessage::new(0, message)?.serialize()?`) — a
+    /// plain `signature.verify(pubkey, message)` over the raw bytes will fail.
+    /// This deviates from the raw-bytes contract of the software backends by
+    /// necessity; see the `sign_message` note on [`SolanaSigner`].
+    async fn sign_message(&self, message: &[u8]) -> Result<Signature, SignerError> {
+        let serialized = solana_offchain_message::OffchainMessage::new(0, message)
```
dev-jodee, line 187: Makes sense that it differs from other because of the above comment, wdyt of putting a helper in this file like 

```
pub fn offchain_envelope(message: &[u8]) -> Result<Vec<u8>, SignerError> {
    solana_offchain_message::OffchainMessage::new(0, message)
        .and_then(|m| m.serialize())
        .map_err(|e| SignerError::ConfigError(format!("invalid off-chain message: {e:?}")))
}
```

and then here you call it, in the lib.rs behind the ledger feature flag we re-export that helper (offchain_envelope), so that callers never need to know about `solana-offchain-message`

rust/src/ledger/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+
+use crate::error::SignerError;
+use crate::sdk_adapter::{Pubkey, Signature, Transaction};
+use crate::traits::{SignTransactionResult, SolanaSigner};
+use crate::transaction_util::TransactionUtil;
+
+/// Default Solana derivation path: `m/44'/501'/0'`.
+///
+/// This matches **Ledger Live**'s Solana accounts (account index, no "change"
+/// component), so the address pay derives equals the one a user sees and funds
```
dev-jodee, line 41: `pay` I think is leftover from the pay kit pr, let's generalize this :)

rust/src/ledger/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    #[test]
+    fn connect_without_device_fails_cleanly() {
+        // Contract: with no usable Ledger, connect returns an error cleanly and
+        // never hangs or panics. We accept any Err (the exact variant depends on
+        // the host's HID subsystem — e.g. NotAvailable when absent, but a CI
+        // runner without libhidapi may surface something else). If a device *is*
+        // attached, connect succeeds and there is nothing to assert.
+        match LedgerSigner::connect(None, false, None) {
```
dev-jodee, line 420: if a ledger is attached when running this unit test it'll actually connect to it right? Might be better to remove the test and keep all real i/o in the integration tests

rust/README.md
```diff
@@ -269,6 +270,42 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 **Note:** Utila `sign_message` is intentionally unsupported in this signer and returns `SigningFailed`. Transaction signing requests are created with `publish=false`; callers remain responsible for broadcasting.
 
+### Ledger Signer
+
+[Ledger](https://www.ledger.com/) hardware-wallet signer over USB-HID, built on
+[`solana-remote-wallet`](https://docs.rs/solana-remote-wallet). The private key
+never leaves the device and every signature must be confirmed on the device
+screen. Requires the `ledger` + `sdk-v3` features, a connected and unlocked
+device running the Solana app (and, on Linux, Ledger's `udev` rules).
```
dev-jodee, line 279: can we add that on linux, since hidapi will compile to native code that users need `pkg-config and libudev-dev` (in case they dont already have it)

dev-jodee (commented):
Thanks for the work ! Couple of comments, also would it be possible to add parity to our typescript package as well?

amilz:
Thanks, @ledgicr -- do you have plans to add TypeScript support?

PR #164, opened 2026-06-18
feat(go): Go implementation foundation — core + memory backend

HealthyBuilder:
## Summary

Adds a **Go** implementation of `solana-keychain` alongside the existing Rust and
TypeScript libraries, with full parity to the shared `SolanaSigner` contract. This
is the **foundation phase** — the shared `core` package plus the `memory` reference
backend. Remote backends, the umbrella dispatcher, CI, and publishing are deferred
to follow-ups (tracked below).

> Draft: opening early for feedback on the layout and contract before the remaining
> backends land.

## What's included

- **`go/core/`** — shared contract (Go analog of `@solana/keychain-core`):
  `Signer` interface (`context.Context`-based), `SignedTransaction`/`Completeness`,
  redacting `SignerError` + stable `Code`s + `SanitizeRemoteResponse`, `txutil`
  (bincode+base64 serialize / signer-position / add-signature / classify),
  `VerifyEd25519`, HTTPS-only `NewHTTPClient`, concurrent batch helpers.
- **`go/signers/memory/`** — in-memory Ed25519 reference backend (base58 /
  `[1,2,…]` u8-array / raw 32|64-byte / Solana CLI keypair file).
- **`go/testutils/`** — deterministic keypair + test-transaction helpers.
- **Dev tooling** — `justfile` `go-*` recipes wired into
  `fmt`/`build`/`test`/`test-integration`; `go/.golangci.yml`; `.gitignore`;
  `.pre-commit-config.yaml`.

## Key decisions

- Single Go module at `go/` (`github.com/solana-foundation/solana-keychain/go`);
  backends grouped under `signers/`. Go's compiler excludes unimported backends —
  the import graph is the "feature flag".
- Built directly on [`gagliardetto/solana-go`](https://github.com/gagliardetto/solana-go);
  **no v2/v3 SDK adapter** (that split doesn't exist in Go).
- Independent versioning: future tag `go/vX.Y.Z`, starting at v0.x.

## Parity verification (the #1 risk)

Transaction serialization is verified **byte-identical to Rust**: `core/parity_test.go`
pins vectors produced by the Rust signer for the same deterministic transaction
(canonical keypair + fixed recipient + zero blockhash + 1,000,000-lamport transfer);
Go's message bytes and full `base64(bincode(tx))` match exactly. Key parsing is also
asserted against the canonical cross-language vector (`TEST_KEYPAIR_BYTES` → `4BuiY9QU…`).

## Testing

[...]

PR #179, opened 2026-07-02
feat(go): remote signer backends to thirteen-backend parity (stacked on #164)

amilz:
## Summary

Builds on #164 (Go foundation: `core` + `memory`) to bring the Go implementation to **full thirteen-backend parity** with Rust and TypeScript:

Memory · Vault · Privy · Turnkey · AWS KMS · Fireblocks · GCP KMS · Dfns · Crossmint · CDP · Para · Openfort · Utila

> **Stacked on #164.** This branch contains #164's commit (rebased onto current `main`) plus the work below; the diff shown here includes both. If #164 merges first, this PR reduces to the backend work.

### Quality pass on the #164 foundation
- 64-byte keypair bytes now validate the embedded public half against the seed-derived key (parity with Rust `Keypair::try_from`)
- `memory.Config.PrivateKey` retyped to `[]byte` — the 32-byte seed form is not a valid `ed25519.PrivateKey` value
- `core.NewHTTPClient` honors `HTTP(S)_PROXY` env config and refuses all redirects (auth headers must never replay against a redirect target)
- Keypair-file error taxonomy: read failures are `IO_ERROR`, malformed contents `INVALID_PRIVATE_KEY`

### Thirteen backends (`go/signers/<pkg>`)
Each is a port of the corresponding Rust module (authoritative spec), with the Rust wiremock unit tests ported to `httptest`/stubbed-client tests — race-clean across all 14 packages:

| Backend | Notable behavior ported |
| --- | --- |
| `vault` | transit sign endpoint, `vault:vN:` prefix stripping, key-metadata health check |
| `turnkey` | P-256 API-key stamping (X-Stamp), **r/s left-padding quirk** + dedicated test |
| `privy` | basic auth + `privy-app-id`, wallet lookup at construction, chain-type guard, authorization-context signing |
| `para` | `sk_`/UUID/HTTPS config validation, hex sign-raw with `0x` handling, 5s-bounded availability |
| `awskms` | `RAW` + `ED25519_SHA_512`, DescribeKey gate (`ECC_NIST_EDWARDS25519`, enabled, `SIGN_VERIFY`) |
| `gcpkms` | **PureEdDSA** — raw `data`, never `Digest`, `EC_SIGN_ED25519` health check |
| `fireblocks` | RS256 request JWTs (uri/nonce/bodyHash), RAW polling; **`UseProgramCall` rejected at construction before any network call** (PROGRAM_CALL broadcasts on-chain — duplicate-spend risk, parity with Rust #153) |
| `dfns` | User Action Signing challenge flow, Ed25519/P-256/RSA credential keys, serde-exact clientData bytes |
| `cdp` | EdDSA bearer + ES256 wallet-auth JWTs with key-sorted `reqHash`, **UTF-8-only `SignMessage`**, tamper rejection |
| `crossmint` | HKDF delegated-signer derivation, `encodeURIComponent`-exact locators, create/poll/approve flow, **`SignMessage` intentionally unsupported** |
| `openfort` | ES256 x-wallet-auth JWTs, dual-format wallet secret (PEM / bare base64 DER) |
| `utila` | poll-to-SIGNED flow with tamper detection against the submitted message |

### Shared conventions across all backends
- `Config` struct + `New` constructor returning a ready-to-use signer; backends with a Rust `init()` take a `context.Context` and initialize inline (analog of Rust `Signer::from_*` / TS async factories)
- HTTPS enforced by default via `core.NewHTTPClient`; optional `HTTPClient` (or SDK-level client) override is the documented escape hatch — the Go analog of Rust `with_client`
- Remote signatures verified locally with `core.VerifyEd25519` before use, wherever Rust does
- Transport-raised `*core.SignerError`s pass through unwrapped in every backend; untrusted remote text passes through `core.SanitizeRemoteResponse`
- Redacting `String()`/`GoString()` on every remote signer; no logging, no globals; signers immutable and concurrency-safe after construction

### Integration & tooling
- Vault integration test (`-tags=integration`) mirroring `test_vault_integration.rs`, wired to the existing `just go-test-integration` recipe
- `go-ci.yml`: build / vet / race-enabled tests / golangci-lint, gated on `go/` changes; `just go-test` runs `-race` to match
- Cross-language parity vectors pinned on both the Go and Rust sides
- `golangci-lint` clean at zero findings (including `--max-same-issues=0`)
- Dependency isolation is structural: **every backend is its own Go module** (15 modules: `core`, `testutils`, 13 signers) — a memory-only consumer's module graph carries no AWS/GCP SDKs, and the `go 1.25.8` toolchain floor forced by `google.golang.org/api` is confined to the `gcpkms` module (everything else is `go 1.25`)

### Per-backend Go modules
In-repo `replace` directives wire the modules together (the cargo path-dep pattern); requires get real versions at first tag, releases tag `go/core` + `go/testutils` before the signer modules. `justfile` go recipes and `go-ci.yml` iterate every `go.mod`. Until the first `go/...` tags exist, `@latest` cannot resolve the in-repo `go/core` requirement (documented in `go/README.md`).

### Umbrella package: intentionally omitted
The TS umbrella and Rust enum stay lean via tree-shaking / cargo features. Go has no dead-code elimination across a runtime dispatch switch, so an umbrella would force the AWS + GCP SDKs into every consumer's build. Importing the backend package is the Go-native selector; rationale documented in `go/README.md`.

### Known follow-up
- #181 — crossmint's serialized-transaction signature verification semantics (deliberate Rust-parity behavior; to be revisited in all three languages together)

## Testing

- `just go-build`, `just go-fmt` (gofmt + vet per module) — clean
- `just go-test` (`go test -race -count=1` per module) — 14/14 test packages ok across all 15 modules
- golangci-lint — 0 issues per module (enforced in CI)
- `go build -tags=integration ./...` — compiles; vault integration runs under `just go-test-integration`

## Deferred

- [ ] Publish workflow (`go/vX.Y.Z`), fork-live-tests for Go
- [ ] `docs/ADDING_SIGNERS.md` Go section + root README three-language presentation
- [ ] Audit-scope decision for Go

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012X3zut1xuLEBvfpGGJfoTT


go/signers/para/signer.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	raw, err := hex.DecodeString(hexStr)
+	if err != nil {
+		return solana.Signature{}, core.WrapSignerError(core.CodeSigningFailed, "failed to decode hex signature", err)
+	}
+	var sig solana.Signature
+	copy(sig[:], raw)
+	return sig, nil
+}
+
```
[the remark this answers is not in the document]
amilz: Fixed in 6c64507 — `doRequest` now passes through transport-raised `*core.SignerError`s (with a regression test) instead of re-wrapping as `CodeHTTPError`, matching the other HTTP backends.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/openfort/signer.go
```diff
@@ -0,0 +1,240 @@
...
+		return solana.Signature{}, core.NewSignerError(core.CodeSigningFailed,
+			"signature verification failed - the returned signature does not match the public key")
+	}
+	return sig, nil
+}
+
+// do executes an Openfort API request and returns the response body of a 2xx
+// response. Transport failures map to CodeHTTPError and non-2xx statuses to
+// CodeRemoteAPIError with only the status code (the Rust error never includes
+// the remote body).
```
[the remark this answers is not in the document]
amilz: Fixed in 6c64507 — `do` now passes through transport-raised `*core.SignerError`s (with a regression test) instead of re-wrapping as `CodeHTTPError`.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/dfns/signer.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return solana.Signature{}, core.WrapSignerError(core.CodeSerializationError, "failed to decode signature r", err)
+	}
+	sBytes, err := hex.DecodeString(strings.TrimPrefix(s, "0x"))
+	if err != nil {
+		return solana.Signature{}, core.WrapSignerError(core.CodeSerializationError, "failed to decode signature s", err)
+	}
+	if len(rBytes)+len(sBytes) != solana.SignatureLength {
+		return solana.Signature{}, core.NewSignerError(core.CodeSigningFailed, "invalid signature length (expected 64 bytes)")
+	}
```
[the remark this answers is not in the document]
amilz: Fixed in 6c64507. Note the Rust `combine_signature` also only checks the 32-byte total (concatenation, no padding), so this was Rust-parity as written — a misaligned split can never pass the downstream `VerifyEd25519`, so nothing invalid was accepted. Tightened Go to require exactly 32-byte r and s anyway for the accurate error message, with a regression test. Left the Turnkey-style left-padding out deliberately: for Dfns, Rust treats component lengths as exact, and padding would diverge from the authoritative implementation.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/core/batch.go
```diff
@@ -0,0 +1,91 @@
...
+func stagger(ctx context.Context, index int, delay time.Duration) error {
+	if delay <= 0 || index == 0 {
+		return nil
+	}
+	timer := time.NewTimer(time.Duration(index) * delay)
+	defer timer.Stop()
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-timer.C:
```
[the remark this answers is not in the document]
amilz: Fixed in 59b258f — `stagger` now wraps cancellation as `CodeHTTPError` with the context error reachable via `errors.Is`, consistent with the crossmint/fireblocks polling sites. Added `core/batch_test.go` covering this plus batch ordering.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/crossmint/signer.go
```diff
@@ -0,0 +1,399 @@
...
+		return solana.Signature{}, core.WrapSignerError(core.CodeSerializationError,
+			"failed to serialize Crossmint transaction message", err)
+	}
+	if err := s.verifySignatureMatchesMessage(signature, remoteMessage); err != nil {
+		return solana.Signature{}, err
+	}
+	return signature, nil
+}
+
+// verifySignatureMatchesMessage checks a remote signature against this wallet's
```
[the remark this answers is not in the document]
amilz: This is a faithful port of deliberate, test-pinned behavior in the authoritative Rust implementation — not a Go-side gap. Rust's `extract_signature_from_serialized_transaction` (rust/src/crossmint/mod.rs) also verifies against the remote transaction's bytes, and two Rust tests pin it explicitly: `test_sign_transaction_accepts_signature_from_on_chain_transaction_bytes` builds a remote transaction with a *different recipient* and asserts the signature over those different bytes is accepted, and `test_sign_transaction_prefers_on_chain_transaction_signature_over_txid_fallback` asserts it wins over the txId path. The Go port reproduces both tests.

Tightening this in Go alone would silently diverge the three implementations for identical Crossmint responses, which is the opposite of what this PR promises. The observation itself is fair — the asymmetry with the txId path (which does check `expectedMessage`) reads as intentional tolerance for Crossmint transaction rewrites, but the resulting signature is then placed into the *original* transaction, so a rewritten response yields an encoded transaction that can't land on-chain without an error. We're tracking that as a cross-language question to resolve in Rust + TS + Go together rather than patching one port: see the follow-up issue.

---
_Generated by [Claude Code](https://claude.ai/code)_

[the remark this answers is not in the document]
amilz: Follow-up filed as #181.

---
_Generated by [Claude Code](https://claude.ai/code)_

amilz:
@greptile-apps review — bypassing the file limit; the latest push only touched `justfile` (`go-test` now runs `-race` to match CI), `CLAUDE.md` (three-language / thirteen-backend overview), and a `//nolint` in the utila redaction test.

---
_Generated by [Claude Code](https://claude.ai/code)_

amilz:
@dev-jodee two notes on the latest rounds:

- Openfort's HTTPS base-URL check is now enforced unconditionally (b84fae5), aligning it with crossmint/para/utila per Greptile's last observation; its tests moved to `httptest.NewTLSServer` to match the pattern you established.
- Greptile's remaining suggestion is in `.github/workflows/go-ci.yml`: the golangci-lint version is pinned (v2.12.2) but the install script is fetched from the mutable `/HEAD/install.sh` URL. This session's token can't modify workflow files, so leaving it to you — pinning the script to the release tag (`.../v2.12.2/install.sh`) or switching to the official `golangci/golangci-lint-action` closes it.

---
_Generated by [Claude Code](https://claude.ai/code)_

amilz:
@greptile-apps review — bypassing the file limit again; since the last full review the branch gained the turnkey `crypto/ecdh` public-key derivation, the crossmint unsigned-wire-format test pin, and the openfort unconditional HTTPS base-URL check (`b84fae5`).

---
_Generated by [Claude Code](https://claude.ai/code)_

dev-jodee (commented):
r

go/core/go.mod
```diff
@@ -0,0 +1,38 @@
+module github.com/solana-foundation/solana-keychain/go/core
+
+go 1.25.0
+
+require (
+	github.com/gagliardetto/solana-go v1.20.0
+	github.com/solana-foundation/solana-keychain/go/testutils v0.0.0-00010101000000-000000000000
```
sonicfromnewyoke: `replace` directives only apply to the main module. 
Once `go/core` is tagged and published, any consumer running `go get github.com/solana-foundation/solana-keychain/go/core` will have the go tool.

tag `go/testutils` first and require a real version, or remove the testutils dependency from core's tests (inline the two small helpers). 

this must be resolved before the module is publishable at all.

go/core/go.mod
```diff
@@ -0,0 +1,38 @@
...
+	go.uber.org/multierr v1.11.0 // indirect
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../testutils
```
sonicfromnewyoke: this one

go/core/errors.go
```diff
@@ -0,0 +1,165 @@
...
+	var se *SignerError
+	if errors.As(err, &se) {
+		return se.Code, true
+	}
+	return "", false
+}
+
+const defaultRemoteErrorResponseMaxLength = 256
+
+var whitespaceRun = regexp.MustCompile(`\s+`)
```
sonicfromnewyoke: i believe it should be `[\s\p{Zs}\x{2028}\x{2029}\x{FEFF}]+` or we have to normalize with `unicode.IsSpace`

go/core/errors.go
```diff
@@ -0,0 +1,165 @@
...
+	var b strings.Builder
+	b.Grow(len(responseText))
+	for _, r := range responseText {
+		if isDisallowedASCIIControl(r) {
+			b.WriteByte(' ')
+		} else {
+			b.WriteRune(r)
+		}
+	}
+	normalized := strings.TrimSpace(whitespaceRun.ReplaceAllString(b.String(), " "))
```
sonicfromnewyoke, line 159: here we are matching only ASCII whitespaces

go/core/batch.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
+
+// stagger sleeps index*delay (respecting ctx) before a task starts, mirroring the
+// TS per-index delay used for rate limiting.
+func stagger(ctx context.Context, index int, delay time.Duration) error {
```
sonicfromnewyoke: stagger sleeps` i x delay` measured from when task i acquires an errgroup slot, not from batch start

 this will silently cause blockhash-expiration failures on large batches - the exact problem the TS >3000ms warning exists for.

go/signers/awskms/go.mod
```diff
@@ -0,0 +1,56 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 54: same here

go/signers/awskms/go.mod
```diff
@@ -0,0 +1,56 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 56: and here

go/signers/cdp/go.mod
```diff
@@ -0,0 +1,42 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 40: and here

go/signers/cdp/go.mod
```diff
@@ -0,0 +1,42 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 42: and 1 more place

go/signers/crossmint/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/multierr v1.11.0 // indirect
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: here

go/signers/crossmint/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and here

go/signers/gcpkms/go.mod
```diff
@@ -0,0 +1,70 @@
...
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/text v0.38.0 // indirect
+	golang.org/x/time v0.15.0 // indirect
+	google.golang.org/genproto v0.0.0-20260319201613-d00831a3d3e7 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20260401024825-9d38bb4040a9 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260622175928-b703f567277d // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 68: 1 more

go/signers/gcpkms/go.mod
```diff
@@ -0,0 +1,70 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 70: and here

go/signers/memory/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: this one

go/signers/memory/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and this

go/signers/openfort/go.mod
```diff
@@ -0,0 +1,42 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 40: here

go/signers/openfort/go.mod
```diff
@@ -0,0 +1,42 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 42: and here

go/signers/para/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: this

go/signers/para/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and this

go/signers/privy/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: 👇

go/signers/privy/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and 👇

go/signers/turnkey/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: 👇

go/signers/turnkey/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and 👇

go/signers/utila/go.mod
```diff
@@ -0,0 +1,42 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 40: ⬇️

go/signers/utila/go.mod
```diff
@@ -0,0 +1,42 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 42: and ⬇️

go/signers/vault/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: ⬇️

go/signers/vault/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and ⬇️

go/signers/fireblocks/go.mod
```diff
@@ -0,0 +1,42 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 40: here

go/signers/fireblocks/go.mod
```diff
@@ -0,0 +1,42 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 42: and here

go/signers/dfns/go.mod
```diff
@@ -0,0 +1,41 @@
...
+	go.uber.org/ratelimit v0.3.1 // indirect
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.53.0 // indirect
+	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/sys v0.46.0 // indirect
+	golang.org/x/term v0.44.0 // indirect
+	golang.org/x/time v0.11.0 // indirect
+)
+
+replace github.com/solana-foundation/solana-keychain/go/core => ../../core
```
sonicfromnewyoke, line 39: this one

go/signers/dfns/go.mod
```diff
@@ -0,0 +1,41 @@
...
+
+replace github.com/solana-foundation/solana-keychain/go/testutils => ../../testutils
```
sonicfromnewyoke, line 41: and this one

go/signers/vault/signer.go
```diff
@@ -0,0 +1,224 @@
...
+func (s *Signer) Pubkey() solana.PublicKey { return s.pubkey }
+
+// String renders the signer without any secret material — the Go analog of the
+// Rust redacting Debug impl.
+func (s *Signer) String() string {
+	return "vault.Signer{pubkey: " + s.pubkey.String() + "}"
+}
+
+// GoString mirrors String so %#v cannot leak secrets either.
+func (s *Signer) GoString() string { return s.String() }
```
sonicfromnewyoke: `String() / GoString()` use a pointer receiver, so the redaction is bypassed

go/signers/memory/signer.go
```diff
@@ -0,0 +1,67 @@
...
+	"crypto/ed25519"
+
+	"github.com/gagliardetto/solana-go"
+
+	"github.com/solana-foundation/solana-keychain/go/core"
+)
+
+// Signer is an in-memory Ed25519 signer. It is the Go analog of the Rust
+// MemorySigner and the TS memory signer.
+type Signer struct {
```
sonicfromnewyoke, line 14: `memory.Signer` has no redacting `String() / GoString()`

 fmt prints the full private key.

go/signers/dfns/auth.go
```diff
@@ -0,0 +1,201 @@
...
+
+// userActionResponse carries the token returned by /auth/action.
+type userActionResponse struct {
+	UserAction string `json:"userAction"`
+}
+
+// clientData is the challenge payload signed by the credential key. The field
+// order (challenge, then type) reproduces the exact byte layout the Rust
+// implementation signs (serde_json's map serialization is key-sorted).
+type clientData struct {
```
sonicfromnewyoke, line 72: rust builds `serde_json::json!({"type": ..., "challenge": ...})`, and serde_json without `preserve_order`.

here is declared Challenge before Type, and `marshalJSON` disables HTML escaping, matching serde's non-escaping of `<, >, &`. The trailing-newline trim is safe (raw `\n` cannot appear inside JSON output)

👍

go/signers/crossmint/derive.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, core.NewSignerError(core.CodeConfigError,
+			fmt.Sprintf("signer_secret must be a 64-char hex string (got %d)", len(rawSecret)))
+	}
+	ikm, err := hex.DecodeString(rawSecret)
+	if err != nil {
+		return nil, core.WrapSignerError(core.CodeConfigError, "signer_secret is not valid hex", err)
+	}
+
+	info := projectID + ":" + environment + ":solana-ed25519"
+	reader := hkdf.New(sha256.New, ikm, []byte(hkdfSalt), []byte(info))
```
sonicfromnewyoke, line 49: it maps exactly to rust `Hkdf::<Sha256>::new(Some(b"crossmint"), &ikm).expand(info)`. 
Info string `{projectId}:{environment}:solana-ed25519`, 64-hex IKM after optional xmsk1_ strip, 32-byte output as Ed25519 seed - all identical.

💪

go/signers/turnkey/signer.go
```diff
@@ -0,0 +1,226 @@
...
+	apiBaseURL     string
+	client         *http.Client
+}
+
+// Ensure Signer satisfies the core contract at compile time.
+var _ core.Signer = (*Signer)(nil)
+
+// New builds a Turnkey signer from cfg. Construction is pure — no network I/O —
+// matching the Rust TurnkeySigner::from_config (Turnkey has no init step).
+func New(cfg Config) (*Signer, error) {
```
sonicfromnewyoke, line 45: HTTPS enforcement is fully bypassable via `Config.HTTPClient`

`APIBaseURL` scheme is never validated.

go/signers/privy/signer.go
```diff
@@ -0,0 +1,251 @@
...
+	authorizationRequestExpiryMs *uint64
+}
+
+// Ensure Signer satisfies the core contract at compile time.
+var _ core.Signer = (*Signer)(nil)
+
+// New builds a Privy signer and fetches the wallet's public key from the Privy
+// API (the Rust init(), performed inline — parity with Signer::from_privy and the
+// TS createPrivySigner factory), so the returned signer is ready to use.
+func New(ctx context.Context, cfg Config) (*Signer, error) {
```
sonicfromnewyoke, line 42: HTTPS enforcement is fully bypassable via `Config.HTTPClient`

`APIBaseURL` scheme is never validated.

go/signers/privy/signer.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	body, err := s.do(req, "privy signing")
+	if err != nil {
+		return solana.Signature{}, err
+	}
+
+	var signResp signMessageResponse
+	if err := json.Unmarshal(body, &signResp); err != nil {
+		return solana.Signature{}, core.WrapSignerError(core.CodeParsingError, "failed to parse privy signing response", err)
```
sonicfromnewyoke: response-JSON parse failures map to `CodeParsingError`, diverging from both the Rust spec and the other Go signers

go/signers/privy/signer.go
```diff
@@ -0,0 +1,251 @@
...
+	req.Header.Set("privy-app-id", s.appID)
+
+	body, err := s.do(req, "privy wallet fetch")
+	if err != nil {
+		return solana.PublicKey{}, err
+	}
+
+	var wallet walletResponse
+	if err := json.Unmarshal(body, &wallet); err != nil {
+		return solana.PublicKey{}, core.WrapSignerError(core.CodeParsingError, "failed to parse privy wallet response", err)
```
sonicfromnewyoke: response-JSON parse failures map to `CodeParsingError`, diverging from both the Rust spec and the other Go signers

go/signers/para/signer.go
```diff
@@ -0,0 +1,292 @@
...
+// fetchWallet retrieves wallet info from GET /v1/wallets/{walletId}. Port of
+// the Rust fetch_wallet.
+func (s *Signer) fetchWallet(ctx context.Context) (*walletResponse, error) {
+	body, err := s.doRequest(ctx, http.MethodGet, s.baseURL+"/v1/wallets/"+s.walletID, nil)
+	if err != nil {
+		return nil, err
+	}
+	var wallet walletResponse
+	if err := json.Unmarshal(body, &wallet); err != nil {
+		return nil, core.WrapSignerError(core.CodeParsingError, "failed to parse para wallet response", err)
```
sonicfromnewyoke: response-JSON parse failures map to `CodeParsingError`, diverging from both the Rust spec and the other Go signers

go/signers/para/signer.go
```diff
@@ -0,0 +1,292 @@
...
+		return solana.Signature{}, core.NewSignerError(core.CodeConfigError, "signer not initialized: call init() first")
+	}
+	request := signRawRequest{Data: hex.EncodeToString(data), Encoding: "hex"}
+	body, err := s.doRequest(ctx, http.MethodPost, s.baseURL+"/v1/wallets/"+s.walletID+"/sign-raw", &request)
+	if err != nil {
+		return solana.Signature{}, err
+	}
+	var response signRawResponse
+	if err := json.Unmarshal(body, &response); err != nil {
+		return solana.Signature{}, core.WrapSignerError(core.CodeParsingError, "failed to parse para sign-raw response", err)
```
sonicfromnewyoke: response-JSON parse failures map to `CodeParsingError`, diverging from both the Rust spec and the other Go signers

go/signers/privy/signer.go
```diff
@@ -0,0 +1,251 @@
...
+	defer func() { _ = resp.Body.Close() }()
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBytes))
+	if err != nil {
+		return nil, core.WrapSignerError(core.CodeHTTPError, "failed to read "+what+" response", err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, core.NewSignerError(core.CodeRemoteAPIError,
+			what+" failed: API error "+strconv.Itoa(resp.StatusCode)+": "+core.SanitizeRemoteResponse(string(body)))
+	}
```
sonicfromnewyoke, line 256: non-2xx Privy responses embed the sanitized remote body in the error detail

 the other three packages (and Rust) deliberately return only the status code.

go/signers/cdp/signer.go
```diff
@@ -0,0 +1,296 @@
...
+func (s *Signer) Pubkey() solana.PublicKey { return s.pubkey }
+
+// String renders the signer without any secret material — the Go analog of the
+// Rust redacting Debug impl.
+func (s *Signer) String() string {
+	return "cdp.Signer{pubkey: " + s.pubkey.String() + ", apiBaseURL: " + s.apiBaseURL + "}"
+}
+
+// GoString mirrors String so %#v cannot leak secrets either.
+func (s *Signer) GoString() string { return s.String() }
```
sonicfromnewyoke: pointer-receiver methods, so they are not in the value type's method set

`fmt.Sprintf("%+v", *s)` falls back to reflective struct printing, which prints unexported fields.
it will print something like `{apiKeyID:test-api-key apiKeySecret:QkJCQkJC...`

go/signers/fireblocks/signer.go
```diff
@@ -0,0 +1,201 @@
...
+
+// String renders the signer without any secret material — the Go analog of the
+// Rust redacting Debug impl.
+func (s *Signer) String() string {
+	return "fireblocks.Signer{pubkey: " + s.pubkey.String() +
+		", vaultAccountID: " + s.vaultAccountID + ", apiBaseURL: " + s.apiBaseURL + "}"
+}
+
+// GoString mirrors String so %#v cannot leak secrets either.
+func (s *Signer) GoString() string { return s.String() }
```
sonicfromnewyoke: same as in CDP -  pointer-receiver methods, so they are not in the value type's method set

go/signers/openfort/signer.go
```diff
@@ -0,0 +1,257 @@
...
+func (s *Signer) Pubkey() solana.PublicKey { return s.pubkey }
+
+// String renders the signer without any secret material — the Go analog of the
+// Rust redacting Debug impl.
+func (s *Signer) String() string {
+	return "openfort.Signer{accountID: " + s.accountID + ", pubkey: " + s.pubkey.String() + ", baseURL: " + s.baseURL + "}"
+}
+
+// GoString mirrors String so %#v cannot leak secrets either.
+func (s *Signer) GoString() string { return s.String() }
```
sonicfromnewyoke: same as in CDP -  pointer-receiver methods, so they are not in the value type's method set

sonicfromnewyoke (changes requested):
Amazing work 😍

2 things i'd like to ask you addressing before the merge:
1. secret leakage through fmt printing - memory.Signer has no redacting `String()/GoString()` at all. `log.Printf("%v", s)` prints the complete 64-byte private key via reflection. Rust's MemorySigner has a manual redacting Debug impl; the Go port just never got one.
vault, cdp, fireblocks, openfort - these do have redacting `String() / GoString()`, but on pointer receivers. Formatting the dereferenced value (`fmt.Sprintf("%+v", *signer)`  - a one-character mistake) bypasses redaction and prints the raw Vault token, CDP API secret, Fireblocks key, or Openfort `sk_live` key.
2.  the modules cannot be published as-is - every go.mod requires its in-repo siblings at the placeholder pseudo-version `v0.0.0-00010101000000-000000000000` plus a local replace. 
Replace directives are ignored when a module is consumed as a dependency, so after tagging, `go get .../go/signers/vault@v0.1.0` will still try to resolve go/core at the placeholder version and fail.

amilz:
thank you @sonicfromnewyoke !

go/core/go.mod
amilz: Addressed in 5e437bf, taking your second option: core's tests no longer touch the testutils module at all (the one helper plus the fixed keypair are inlined in `core/txutil_helpers_test.go`), so `go/core` has zero in-repo dependencies — not even test-only — and no `replace` directives; it is taggable and publishable first, unconditionally.

For the thirteen signer modules the `replace` directives remain a development-time device (there is nothing on the proxy to require yet). What makes them publishable is the release sequence, now mechanized: `just go-release-prep vX.Y.Z` drops the replaces and pins the real released versions in every signer go.mod after `go/core` and `go/testutils` are tagged, and `go/README.md` documents the three-step order (tag core+testutils → release-prep + commit → tag signers). Your other line-pointers on the signer go.mod files are all this same pattern, so I've resolved them against this thread.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/core/errors.go
amilz: Good catch — Go's `\s` is ASCII-only while the JS `\s` this ports is Unicode-aware. Fixed in 5e437bf with exactly your class: `[\s\p{Zs}\x{2028}\x{2029}\x{FEFF}]+` (which is byte-for-byte the JS `\s` set), with new test cases covering NBSP/figure/thin/ideographic spaces, U+2028/U+2029, BOM, and the unicode-whitespace-only → `[empty remote response]` path. This is the same regex used at both flagged sites (the line-155 use is the line-125 definition).

---
_Generated by [Claude Code](https://claude.ai/code)_

go/core/batch.go
amilz: Fixed in 5e437bf — `stagger` now targets `batchStart + i*delay` and sleeps only the remaining time (a task whose slot opened past its target starts immediately), so bounded-concurrency batches can't compound slot-waits with per-index sleeps into blockhash expiry. Pinned by `TestStaggerAnchorsToBatchStart` (a task with a long-past target must not sleep).

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/vault/signer.go
amilz: Right — with pointer receivers, printing the dereferenced value (`%+v` on `*s`) falls back to reflective struct printing and dumps the unexported fields. Fixed in 5e437bf: `String()`/`GoString()` moved to value receivers in **all ten** remote signers (this thread plus your cdp/fireblocks/openfort ones are the same class), so both the value and pointer method sets carry the redacting formatters. Every redaction test now renders `%v`/`%+v`/`%s`/`%#v` for both `s` and `*s`.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/memory/signer.go
amilz: Confirmed — this was the worst of the class: no formatters at all, so any fmt verb reflectively printed the raw `ed25519.PrivateKey` bytes. Fixed in 5e437bf: `memory.Signer` gains value-receiver redacting `String()`/`GoString()` (type + pubkey only), and a new `memory/redaction_test.go` renders all four verbs on both `s` and `*s` and asserts no private-key material (decimal byte runs or base64 fragments) appears.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/turnkey/signer.go
amilz: Fixed in 5e437bf — and swept the whole tree, since this wasn't just turnkey/privy: dfns, fireblocks, and cdp had the same gap. All five now reject a non-HTTPS `APIBaseURL` at construction with `SIGNER_CONFIG_ERROR`, regardless of whether a custom `HTTPClient` is supplied (matching crossmint/para/utila/openfort, which already enforced unconditionally). Tests moved to `httptest.NewTLSServer`, and each package's HTTPS test asserts the rejection for both the default and a custom client.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/privy/signer.go
amilz: Fixed in 5e437bf — all four privy/para response-parse sites now map to `SIGNER_SERIALIZATION_ERROR`, matching the Rust `serde_json` → `SignerError::SerializationError` conversion and the other Go backends; test expectations updated to pin the new code.

---
_Generated by [Claude Code](https://claude.ai/code)_

go/signers/privy/signer.go
amilz: Fixed in 5e437bf — privy's non-2xx errors now carry only `"… failed: API error <status>"`, matching the Rust implementation (which reads the body into `_error_text` and deliberately discards it outside `unsafe-debug`) and the other Go backends. The former sanitized-body test now pins the opposite: `TestSignMessageRemoteErrorBodyDiscarded` asserts the body never reaches the detail.

---
_Generated by [Claude Code](https://claude.ai/code)_

amilz:
@greptile-apps review — bypassing the file limit; the latest push (5e437bf) addresses @sonicfromnewyoke's review: value-receiver redacting formatters everywhere (+ memory), Unicode-aware whitespace in `SanitizeRemoteResponse`, batch stagger anchored to batch start, unconditional HTTPS base-URL checks in turnkey/privy/dfns/fireblocks/cdp, privy/para error-taxonomy alignment with Rust, and `go/core` freed of its testutils dependency (plus `just go-release-prep` + a documented tag sequence for the signer modules).

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/go-ci.yml
```diff
@@ -0,0 +1,236 @@
...
+  go-lint:
+    needs: resolve-signers
+    if: ${{ needs.resolve-signers.outputs.go_unit_module_count != '0' }}
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: go
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-go@v7
```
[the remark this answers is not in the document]
amilz: Agreed this is worth closing. This session's App token can't modify workflow files, so the change needs to come from a maintainer — @dev-jodee, two equivalent options: pin the installer script to the release tag (`https://raw.githubusercontent.com/golangci/golangci-lint/v2.12.2/install.sh`) or a commit SHA, or switch the step to the official `golangci/golangci-lint-action` pinned like the other actions in this workflow. Leaving this thread open until that lands.

---
_Generated by [Claude Code](https://claude.ai/code)_

lgalabru (approved):

dev-jodee (approved):

dev-jodee:
thanks a lot @sonicfromnewyoke for the review ! all fixes implemented :)

PR #211, opened 2026-08-04
feat(python): GCP KMS signer backend

dev-jodee:
## Summary
- Adds the **GCP KMS** backend (`solana_keychain.gcp_kms`). Stacked on #210 — this layer shows only the GCP work.
- Signing: `AsymmetricSign` in PureEdDSA mode — the request carries the **raw data, never a digest**; returned signatures are length-checked (64 bytes) and verified locally against the configured pubkey before being accepted.
- Availability: `GetPublicKey` gate — the key version's algorithm must be `EC_SIGN_ED25519`.
- Config: `GcpKmsSignerConfig(key_name, public_key, client=None)` — `key_name` is the full `cryptoKeyVersions` resource path; pubkey validated at construction before any GCP call; optional pre-configured `KeyManagementServiceAsyncClient` for custom credentials/endpoint/transport. Async `create_gcp_kms_signer()` factory (client stays on the event loop — the gRPC async transport binds to the running loop, so construction must not move to a worker thread).
- `google-cloud-kms` behind the `[gcp-kms]` extra; importing without the extra raises an `ImportError` naming it. Native async client — no thread bridging needed for calls.

## Test Plan
- 144 tests pass (13 new): raw-data request shape asserted, API error, empty/wrong-length signature, verification failure against mismatched pubkey, transaction signing, algorithm gate, availability error path, pubkey rejected before any GCP call, repr contents

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -25,7 +25,7 @@ library offers a consistent API across all signing methods.
 | **Turnkey** | Non-custodial key management via Turnkey | — | Planned |
 | **AWS KMS** | AWS Key Management Service with Ed25519 signing | `solana_keychain.aws_kms` | ✅ Available |
 | **Fireblocks** | Fireblocks institutional custody platform | — | Planned |
-| **GCP KMS** | Google Cloud Key Management Service with Ed25519 signing | — | Planned |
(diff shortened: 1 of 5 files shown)


amilz (approved):
GCP KMS Ed25519 signer backend; verifies signature against pubkey, full CI green.

PR #215, opened 2026-08-04
feat(python): Para signer backend

dev-jodee:
## Summary
- Adds the **Para** backend (`solana_keychain.para`). Stacked on #214 — this layer shows only the Para work.
- **Construction validation**: `api_key` must carry the `sk_` secret-key prefix, `wallet_id` must be a canonical dashed UUID (8-4-4-4-12), base URL HTTPS-enforced — all `SIGNER_CONFIG_ERROR` before any network call.
- **Init-required**: `init()` (awaited by `create_para_signer`) fetches the wallet with `X-API-Key`, requires `type == SOLANA` (case-insensitive), warns when status isn't `ACTIVE`/`READY`, and resolves the address into the pubkey. Pre-init signing raises `SIGNER_NOT_INITIALIZED`.
- **Signing**: `sign-raw` with hex-encoded data; responses accept optional `0x` prefixes and must decode to exactly 64 bytes; signatures verified locally against the wallet pubkey.
- **Availability**: wallet fetch bounded to 5 seconds (`asyncio.wait_for`), gating on wallet type and `ACTIVE`/`READY` status.
- Pure-httpx backend — no optional extra; exported from the package root like memory/vault. Requests use a 30s timeout.

## Test Plan
- 224 tests pass (32 new): config validation matrix (prefix, UUID shapes, empties, HTTP URL), wallet-type gate incl. case-insensitivity, status warning, missing/invalid address, `0x`-prefixed signatures, wrong-length/undecodable signatures, verification failure, API errors, availability matrix (type/status combinations), real timeout path (patched window + slow mock), not-initialized paths, factory init, repr redaction

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -27,7 +27,7 @@ library offers a consistent API across all signing methods.
 | **Fireblocks** | Fireblocks institutional custody platform | — | Planned |
 | **GCP KMS** | Google Cloud Key Management Service with Ed25519 signing | `solana_keychain.gcp_kms` | ✅ Available |
 | **Dfns** | Dfns wallet infrastructure with Ed25519 signing | — | Planned |
-| **Para** | MPC wallets with Para infrastructure | — | Planned |
+| **Para** | MPC wallets with Para infrastructure | `solana_keychain.para` | ✅ Available |
 | **CDP** | Coinbase Developer Platform managed wallets | — | Planned |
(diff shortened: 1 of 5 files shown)


amilz (approved):
Para signer backend; api_key redacted, verifies returned signature against pubkey, full CI green.

PR #216, opened 2026-08-04
feat(python): Fireblocks signer backend

dev-jodee:
## Summary
- Adds the **Fireblocks** backend (`solana_keychain.fireblocks`). Stacked on #215 — this layer shows only the Fireblocks work.
- **RAW signing with polling**: `POST /v1/transactions` (RAW operation, hex message content, VAULT_ACCOUNT source) then poll `GET /v1/transactions/{id}` until `COMPLETED`; `FAILED`/`CANCELLED`/`REJECTED`/`BLOCKED` are terminal `SIGNING_FAILED`; attempt cap surfaces a polling-timeout `REMOTE_API_ERROR` noting the request may still complete. Interval (default 1s) and attempts (default 300) configurable.
- **RS256 request JWTs**: claims carry `uri`, a fresh UUID `nonce`, `iat`/`nbf` backdated by a 60s skew leeway, 120s TTL, `sub = api_key`, and `bodyHash` = SHA-256 of the exact body bytes — sent via `fetch_signer_json(content=…)` so stamped and wire bytes are identical. Headers: `X-API-Key` + `Authorization: Bearer`.
- **`use_program_call` rejected at `init()` before any network call** (`CONFIG_ERROR`): PROGRAM_CALL broadcasts the transaction on-chain and returns no reusable signer-bound signature — a duplicate-spend risk that violates the signing contract. The signer always uses RAW.
- Signature extraction from `signedMessages[0].signature.fullSig`: hex-decoded, 64-byte length check, verified locally against the vault pubkey. Init resolves the pubkey from `addresses_paginated`. `cryptography` + `pyjwt` behind the `[fireblocks]` extra.

## Test Plan
- 249 tests pass (25 new): JWT claims shape (iat==nbf, exp window, bodyHash) and cryptographic verification of the Bearer token against captured request bytes, PROGRAM_CALL rejection with zero network calls asserted, polling through intermediate statuses, all four terminal statuses, polling timeout, missing/undecodable/wrong-length signatures, verification failure, empty/invalid addresses, RSA-key failure surfaced on use, not-initialized paths, availability paths, factory init, repr redaction (API key + PEM)
- `ruff` + `mypy --strict` + sdist/wheel build clean

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #217, opened 2026-08-04
feat(python): CDP signer backend

dev-jodee:
## Summary
- Adds the **CDP** backend (`solana_keychain.cdp`). Stacked on #216 — this layer shows only the CDP work.
- **Dual JWT auth**: every request carries an EdDSA bearer token (protected header: `kid` = API key id, fresh `nonce`; claims: `sub`, `iss=cdp`, 120s window, `uris = ["METHOD host/path"]`). Write endpoints add an ES256 `X-Wallet-Auth` token whose `reqHash` is SHA-256 over the request body serialized with recursively sorted keys — omitted for empty bodies. API key secret is a validated base64 Ed25519 keypair (pubkey half must match the seed-derived key; PEM keys rejected); wallet secret is base64 PKCS8 DER P-256.
- **`sign_message` is UTF-8-only** — the endpoint takes a string, so non-UTF-8 bytes raise `SIGNER_SERIALIZATION_ERROR`. Returned signatures are base58, length-checked, and verified locally.
- **`sign_transaction` flow differs from other backends**: submits the full base64 serialized transaction, decodes the returned signed transaction, extracts this signer's signature at its required-signer position, verifies it against the message bytes, then applies it to the caller's transaction. Non-signer transactions fail before any network call.
- Availability: bearer-only GET of the account resource. No init needed — the address is supplied in config and validated at construction. `cryptography` + `pyjwt` behind the `[cdp]` extra.

## Test Plan
- 280 tests pass (31 new): both JWTs cryptographically verified against captured requests (kid/nonce/uris/reqHash), reqHash key-order independence and empty-body omission, invalid API/wallet secrets (PEM, bad base64, wrong length, mismatched pubkey half, non-EC DER), UTF-8 rejection, base58/length/verification failures, full signed-transaction round-trip incl. undecodable/undeserializable/unsigned responses, non-signer fast-fail with zero network calls asserted, bearer-only availability, empty-config matrix, repr redaction

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -28,7 +28,7 @@ library offers a consistent API across all signing methods.
 | **GCP KMS** | Google Cloud Key Management Service with Ed25519 signing | `solana_keychain.gcp_kms` | ✅ Available |
 | **Dfns** | Dfns wallet infrastructure with Ed25519 signing | — | Planned |
 | **Para** | MPC wallets with Para infrastructure | `solana_keychain.para` | ✅ Available |
-| **CDP** | Coinbase Developer Platform managed wallets | — | Planned |
+| **CDP** | Coinbase Developer Platform managed wallets | `solana_keychain.cdp` | ✅ Available |
 | **Crossmint** | Crossmint managed wallets | — | Planned |
 | **Openfort** | Openfort backend wallets with TEE-stored keys | — | Planned |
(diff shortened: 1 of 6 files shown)


amilz (approved):
CDP signer backend; EdDSA+ES256 JWT auth, verifies pubkey matches seed and signature against pubkey, full CI green.

PR #223, opened 2026-08-05
test(python): drop tests that cannot fail, tighten the rest

dev-jodee:
## Summary
An audit of all 19 Python test files for tests with no failure mode a user could hit. Net: **−88 lines, 6 tests removed, 8 simplified, 1 strengthened** — no behavioral coverage lost.

**Removed (provably cannot fail):**
- `test_aws_kms_signer.py::test_key_id_variations_accepted` — `key_id` is stored verbatim with no validation, so all three params asserted a getter returns the string just passed in
- `test_aws_kms_signer.py::test_repr_shows_key_id_pubkey_region_only` — the config holds no secret, so this pinned formatting down to `region=None` with no security property behind it
- `test_keychain.py::test_dispatches_vault_backend` — the umbrella has no per-backend branching, so it re-ran the memory dispatch path (its `@respx.mock` was dead too, as no HTTP occurs)
- `test_errors.py::test_copy_and_deepcopy_work` — `copy`/`deepcopy` route through the same `__reduce__` the pickle test already asserts
- `test_crossmint_signer.py::test_signer_secret_default_locator` — poked private `_signer`; the awaiting-approval test already asserts that exact locator on the wire
- `test_memory_signer.py::test_is_available` — the method is `return True`

**Simplified:** collapsed parametrize cases sharing one code path (redaction is code-independent: 13 → 1; three malformed-URL inputs and one UUID input hit the same branch), dropped a dead `signature=None` parameter in the Dfns status test, and replaced exact-repr equality asserts with the secret-absence checks that were the actual intent.

**Strengthened:** `test_memory_signer.py::test_sign_message` now verifies the signature against the pubkey instead of only checking it is 64 bytes — a well-formed wrong signature previously passed. Every remote backend's test already verified properly.

Deliberately untouched: signature verification, redaction/leak checks, error-code taxonomy, provider quirks (r/s left-padding, UTF-8-only messages, `0x` prefixes, unsupported `sign_message`, v0 `0x80` prefix), golden wire-format vectors, and HTTPS/key-material validation.

## Test Plan
- 401 unit tests pass (was 419; the delta is the removed and collapsed cases)
- `ruff` + `mypy --strict` clean
- Spot-checked that each quirk test survives: vault prefix stripping, Turnkey trimmed-component padding, CDP UTF-8 rejection, Crossmint unsupported `sign_message`, golden vectors

### DIFF
--- a/python/tests/test_aws_kms_signer.py
+++ b/python/tests/test_aws_kms_signer.py
@@ -58,32 +58,6 @@ def test_invalid_pubkey_rejected_before_any_aws_call(invalid: str) -> None:
     assert excinfo.value.code == SignerErrorCode.INVALID_PUBLIC_KEY
 
 
-@pytest.mark.parametrize(
-    "key_id",
-    [
-        TEST_KEY_ID,
-        "12345678-1234-1234-1234-123456789012",
-        "alias/my-key",
-    ],
-)
-def test_key_id_variations_accepted(key_id: str) -> None:
-    keypair = Keypair()
-    signer = AwsKmsSigner(
-        AwsKmsSignerConfig(key_id=key_id, public_key=str(keypair.pubkey()), region=TEST_REGION)
-    )
-    assert signer.key_id == key_id
-    assert signer.pubkey == keypair.pubkey()
-
-
-def test_repr_shows_key_id_pubkey_region_only() -> None:
-    keypair = Keypair()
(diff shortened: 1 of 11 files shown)


amilz (approved):

amilz (approved):
Test cleanup: drops tautological/duplicate tests, tightens memory signer test to verify signatures. CI green. Approved (note: PR already merged).

PR #231, opened 2026-08-11
fix(crossmint): make managed-broadcast signing work and report it honestly

dev-jodee:
Verified end-to-end against the live Crossmint API in all four languages. Single commit, based on `main`.

## What the live API actually does

A 69-byte legacy message submitted comes back as a **321-byte v0 message with three required signers**. Only slot 0 is filled, and that signature belongs to **Crossmint's own fee payer** — it sponsors gas, so it becomes the fee payer and rewrites the message before signing. `onChain.txId` is that same fee-payer signature.

**This wallet's signature is in `approvals.submitted[]`**, covering the rewritten message, which is byte-identical to the approval challenge.

### 1. Signing did not work at all for a smart wallet

I confirmed this by running the live suite against `main`: it **fails**. Every port looked for the signature in the returned transaction's signer slots, where it never appears.

The live suites only passed because they reached past the public API to overwrite the signer's private identity field from `TEST_CROSSMINT_SIGNER_DERIVED_PUBKEY`. So the shipped configuration worked only if the caller patched internals, and the production path was exercised solely in a shape the tests themselves acknowledged was not the real one.

Signatures are now located by matching `approvals.submitted`, or a signature slot, against the keys configuration makes known — the wallet address for an `mpc` wallet, plus any delegated signer derivable from `signerSecret` or a `server:` locator — and verified locally against the bytes Crossmint executed. A key that is neither is still rejected. The test override is deleted, along with the Rust test-only field visibility it required.

### 2. The result was misreported

A signature covering Crossmint's bytes was inserted into the **caller's** transaction, serialized, and returned as complete. Those bytes can never verify, so anyone broadcasting the result got a rejection while the library reported success.

When the returned message differs from the submitted one, the result is now a broadcast result: the signature alone, an empty encoded transaction, the caller's transaction untouched, and complete — Crossmint has already landed it and nothing remains for the caller to send. When the messages match, the signature does cover the caller's bytes and is still placed in the transaction.

Note the comparison **decides how to report; it never refuses the provider's answer.** That is the distinction from the closed #230.

## ⚠️ Breaking change (TypeScript)

Crossmint always executes server-side, so it now implements `TransactionSendingSigner`:

```typescript
const [signature] = await signer.signAndSendTransactions([transaction]);
```

`signTransactions()` rejects with `SIGNER_CONFIG_ERROR`. A `SignatureDictionary` keyed to the signer's address asserts the signature covers the caller's `messageBytes`, which under sponsorship it never does, so there was no honest value to return. New exported type: `CrossmintSendingSigner`.

Python, Rust and Go keep one `sign_transaction`, since their return type expresses both outcomes.

## Smaller items

- An abort now stops the polling loop instead of running to the attempt budget. The method documents that aborting cannot recall work Crossmint has accepted.
- A verification failure no longer falls through to `txId` silently. That path only succeeds when Crossmint signed the caller's exact bytes, so for a rewritten transaction it is not a real fallback and its error explains nothing.
- `signerSecret` now documents what the approval challenge **is** — the message of the transaction Crossmint will execute, not derivable from the one submitted — plus the property the signer verifies after the fact (the approval it submitted covers the transaction that executed) and the one it cannot (that the transaction matches caller intent).

## For reviewers

Three code paths are dead for a gas-sponsored smart wallet but kept deliberately: the signature-slot scan, the `txId` verification, and the "returned == submitted, so attach to the caller's transaction" branch. All three are the **mpc / unsponsored** path, which `init()` explicitly accepts. I had no mpc wallet to test against, so removing them would be guesswork in the other direction. Worth a second opinion.

## Testing

Live API, all four languages, run locally:

- `pytest -m integration -k crossmint` — 2 passed
- `go test -tags integration` — ok
- `cargo test --features integration-tests tests::test_crossmint_integration::` — 3 passed
- `vitest` integration — 3 passed

Unit coverage added for the shape the live API returns, so this is covered without credentials: signature sourced from `approvals.submitted`; an approval from an unconfigured key rejected; a delegated-signer signature located and verified; the wallet address occupying a slot it never signed not shadowing the delegated signer; a `txId` signed by the delegated signer accepted; a rewritten response reported as a broadcast result with the caller's transaction left unsigned. Each was confirmed to fail with its change removed.

Full local suites: Python 457 + ruff + mypy strict; Rust on each of `sdk-v2`/`sdk-v3`/`sdk-v4` plus clippy `-D warnings --all-targets` with `integration-tests`; Go `vet -tags integration` and `test -race` across every module; TypeScript build, typecheck, eslint, prettier, tests and treeshake.

### DIFF
--- a/typescript/packages/crossmint/src/crossmint-signer.ts
+++ b/typescript/packages/crossmint/src/crossmint-signer.ts
@@ -8,6 +8,7 @@ import {
     fetchSignerJson,
     normalizeBaseUrl,
     sanitizeRemoteErrorResponse,
+    SignerError,
     SignerErrorCode,
     SolanaSigner,
     throwSignerError,
@@ -66,10 +67,11 @@ let base64Encoder: ReturnType<typeof getBase64Encoder> | undefined;
 /**
  * Crossmint is a broadcast-managed signer: it rewrites the transaction (gas
  * sponsorship, priority fee, its own blockhash) and broadcasts server-side, so
- * returned signatures cover Crossmint's bytes, not the caller's `messageBytes`.
- * The signature is a send result identifying what Crossmint landed, which is why
- * this signer implements {@link TransactionSendingSigner} instead of returning
- * signature dictionaries keyed to the caller's message.
+ * its signatures cover Crossmint's bytes, not the caller's `messageBytes`.
+ * The returned value is the landed transaction's fee-payer signature, the
+ * identifier Solana RPC looks it up by, which is why this signer implements
+ * {@link TransactionSendingSigner} instead of returning signature dictionaries
+ * keyed to the caller's message.
  */
 class CrossmintSigner<TAddress extends string = string> implements CrossmintSendingSigner<TAddress> {
     readonly address: Address<TAddress>;
@@ -432,11 +434,8 @@ class CrossmintSigner<TAddress extends string = string> implements CrossmintSend
         let embeddedError: unknown;
         if (response.onChain?.transaction) {
             try {
-                return await this.extractSignatureFromSerializedTransaction(response.onChain.transaction);
+                return await this.transactionIdFromExecuted(response, response.onChain.transaction);
             } catch (error) {
-                // A rewritten transaction's signature is in approvals.submitted.
-                const approved = await this.signatureFromApprovals(response, response.onChain.transaction);
-                if (approved) return approved;
                 // Keep this error as the cause: it names the check that failed,
                 // where the txId path only reports a mismatch.
                 embeddedError = error;
@@ -460,12 +459,70 @@ class CrossmintSigner<TAddress extends string = string> implements CrossmintSend
                     lastError = error;
                 }
             }
-            throw embeddedError ?? lastError;
+            const cause = embeddedError ?? lastError;
+            if (cause instanceof SignerError) throw cause;
+            throwSignerError(SignerErrorCode.SIGNING_FAILED, {
+                address: this.address,
+                cause,
+                message: 'Crossmint returned no signature verifiable by a configured signer',
+            });
         }
 
         throwSignerError(SignerErrorCode.SIGNING_FAILED, {
+            cause: embeddedError,
             message: 'Unable to extract signature from Crossmint transaction response',
         });
+    }
+
+    /**
+     * Identifier of the transaction Crossmint landed: its fee-payer (slot 0)
+     * signature, the value Solana RPC addresses transactions by, so the caller
+     * can pass the result to `getTransaction`/`confirmTransaction`. Under gas
+     * sponsorship the fee payer is Crossmint's sponsor key, not this wallet.
+     *
+     * Returned only after a configured signer's signature (in a slot or in
+     * `approvals.submitted`) verifies over the executed bytes, proving this
+     * wallet took part in what landed.
+     */
+    private async transactionIdFromExecuted(
+        response: CrossmintTransactionResponse,
+        serializedTransaction: string,
+    ): Promise<SignatureBytes> {
+        base58Encoder ||= getBase58Encoder();
+        const decoded = getTransactionDecoder().decode(base58Encoder.encode(serializedTransaction));
+
+        const participant =
+            (await this.verifiedSignatureFromSlots(decoded)) ??
+            (await this.verifiedSignatureFromApprovals(response, decoded.messageBytes));
+        if (!participant) {
+            throwSignerError(SignerErrorCode.SIGNING_FAILED, {
+                address: this.address,
+                message: 'No configured signer holds a verifying signature in the Crossmint transaction',
+            });
+        }
+
+        const [feePayer, feePayerSignature] = Object.entries(decoded.signatures)[0] ?? [];
+        if (participant.address === feePayer) {
+            return participant.signature;
+        }
+        if (!feePayerSignature) {
+            throwSignerError(SignerErrorCode.SIGNING_FAILED, {
+                message: 'Crossmint transaction carries no fee-payer signature to identify it by',
+            });
+        }
+        try {
+            await assertSignatureValid({
+                data: decoded.messageBytes,
+                signature: feePayerSignature,
+                signerAddress: feePayer as Address,
+            });
+        } catch (error) {
+            throwSignerError(SignerErrorCode.SIGNING_FAILED, {
+                cause: error,
+                message: 'Crossmint fee-payer signature does not verify over the executed transaction',
+            });
+        }
+        return feePayerSignature;
     }
 
     /**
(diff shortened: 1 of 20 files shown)


typescript/packages/crossmint/src/crossmint-signer.ts
```diff
@@ -176,22 +195,37 @@ class CrossmintSigner<TAddress extends string = string> implements SolanaSigner<
...
-        const results: SignatureDictionary[] = [];
+        const results: SignatureBytes[] = [];
         for (const [index, transaction] of transactions.entries()) {
             if (this.requestDelayMs > 0 && index > 0) {
                 await new Promise(resolve => setTimeout(resolve, this.requestDelayMs));
             }
-            const signature = await this.signTransactionManaged(transaction);
-            results.push(createSignatureDictionary({ signature, signerAddress: this.address }));
+            config?.abortSignal?.throwIfAborted();
+            results.push(await this.signTransactionManaged(transaction));
```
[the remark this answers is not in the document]
dev-jodee: Valid, fixed in b97faef. `abortSignal` is now threaded into `signTransactionManaged` and checked at the top of each polling iteration and again after the poll delay, so a cancelled call stops waiting instead of running to `maxPollAttempts`.

Worth being explicit about the limit, which the method now documents: aborting stops *this client* from waiting. It cannot recall a transaction Crossmint has already accepted, so an aborted transaction may still land server-side. There is no cancel endpoint to call.

Test: `stops polling when the abort signal fires mid-flight` aborts from inside a poll response and asserts the fetch count stays far below the 50-attempt budget. Confirmed to fail with the signal checks removed.

typescript/packages/crossmint/src/crossmint-signer.ts
```diff
@@ -365,53 +429,126 @@ class CrossmintSigner<TAddress extends string = string> implements SolanaSigner<
...
             try {
                 return await this.extractSignatureFromSerializedTransaction(response.onChain.transaction);
-            } catch {
-                // If Crossmint returned an onChain.transaction but it could not be
-                // decoded or validated, fall through to txId and still require
-                // cryptographic validation against the original message bytes.
+            } catch (error) {
+                // A rewritten transaction's signature is in approvals.submitted.
+                const approved = await this.signatureFromApprovals(response, response.onChain.transaction);
+                if (approved) return approved;
```
amilz: **The returned value is not the landed transaction's identifier.** On the sponsored-smart-wallet path (the primary one), this returns the wallet's *approval* signature over the rewritten message. Kit's `TransactionSendingSigner` contract — and Solana RPC — treat the returned `SignatureBytes` as the transaction ID, i.e. the slot-0 fee-payer signature, which under sponsorship is Crossmint's sponsor key and is available as `onChain.txId`.

Concretely: `const [sig] = await signer.signAndSendTransactions([tx])` followed by `rpc.getTransaction(sig)` returns `null` even though the transaction landed, because a non-fee-payer signature cannot address a transaction. The README added in this PR compounds it by describing the value as the signature identifying the transaction Crossmint landed.

The approval signature is the right thing to *verify* (proof our key approved the executed bytes) but the wrong thing to *return*. Suggest: keep this verification exactly as is, then return the verified `onChain.txId` / slot-0 signature of the executed transaction so the result is actually addressable on chain.

Worth cross-checking whether the Rust/Go/Python `SignTransactionResult` modeling in this PR returns the same approval signature as the "signature" — and how that interacts with #241.

typescript/packages/crossmint/src/crossmint-signer.ts
```diff
@@ -365,53 +429,126 @@ class CrossmintSigner<TAddress extends string = string> implements SolanaSigner<
...
+                        data: transaction.messageBytes,
+                        signature: fromTxId,
+                        signerAddress: candidate,
+                    });
+                    return fromTxId;
+                } catch (error) {
+                    lastError = error;
+                }
+            }
+            throw embeddedError ?? lastError;
```
amilz: **This can reject with a raw non-`SignerError`.** `embeddedError` is the transaction decoder's unwrapped `Error` (and `lastError` comes from `assertSignatureValid`), so callers branching on the documented stable `error.code` values get an unclassified error here — every other failure path in the package throws `SignerError`. One of the updated unit tests pins the raw behavior (`rejects.toThrow('decode failed')`).

Suggest wrapping in `SIGNING_FAILED` with `cause: embeddedError ?? lastError` — keeps the improved diagnostics without breaking the error contract, and keeps raw provider-derived strings out of top-level messages per the sanitization convention.

amilz (commented):
TypeScript-only review (Rust/Go/Python out of scope here). The signature-location logic itself checks out: candidates are strictly configuration-derived, every accepted signature is verified against the bytes Crossmint executed, unknown keys are still rejected, and the test-override hack is genuinely gone. Two findings below should be resolved before merge; everything shape-related (shared sending-signer type, method removal vs throwing, umbrella/kit-plugin wiring, abort forwarding, locator fallback) is intentionally deferred to DEV-906.

typescript/packages/crossmint/src/crossmint-signer.ts
dev-jodee: Fixed in 04dd8cb. Cross-check confirmed: Rust/Go/Python returned the approval signature too, so all four languages now return the executed transaction's verified slot-0 fee-payer signature (equals `onChain.txId` on the sponsored path). The approval still gates acceptance. One new test per language pins it: sponsor signed at slot 0, wallet approval in `approvals.submitted`, fee-payer signature returned, approval not.

Side effect: a success response without a verifying fee-payer signature now errors; the Crossmint follow-up to #241 will classify that as `BROADCAST_UNCONFIRMED`.

typescript/packages/crossmint/src/crossmint-signer.ts
dev-jodee: Fixed in 04dd8cb. Both escapes are wrapped in `SIGNING_FAILED` with the original as `cause`; a cause that is already a `SignerError` is rethrown as-is. The pinned test now asserts the stable code plus cause.

typescript/packages/crossmint/src/crossmint-signer.ts
```diff
@@ -205,13 +264,19 @@ class CrossmintSigner<TAddress extends string = string> implements SolanaSigner<
...
 
+    /**
+     * `abortSignal` stops this client from waiting; it cannot recall work Crossmint
+     * has already accepted, so an aborted transaction may still land server-side.
+     */
     private async signTransactionManaged(
         transaction: Transaction & TransactionWithinSizeLimit & TransactionWithLifetime,
+        abortSignal?: AbortSignal,
     ): Promise<SignatureBytes> {
         let response = await this.createTransaction(transaction);
```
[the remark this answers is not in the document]
dev-jodee: Addressed in #245 (stacked on this PR): the abort signal is forwarded into every create/get/approval request via `AbortSignal.any` (preserving the default fetch timeout), and the poll and `requestDelayMs` sleeps are abort-aware. Kept there rather than here so the stack stays conflict-free.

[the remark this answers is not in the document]
dev-jodee: Duplicate of the thread above: addressed in #245 (stacked on this PR) via `AbortSignal.any` request forwarding and abort-aware sleeps.

amilz (approved):

dev-jodee:
Superseded by #269, which carries this branch's commits unchanged (this branch is an ancestor of #269's tip). Closing to land the stack as a single squash into `main`. Branch kept for now.

PR #247, opened 2026-08-18
fix(deps): bump h2 to 0.4.16 for RUSTSEC-2026-0258

dev-jodee:
## Summary
- The nightly `Security` workflow started failing on `main` after RUSTSEC-2026-0258 was published on 2026-08-17: h2 below 0.4.16 accepts and queues empty DATA frames without limit, which can grow memory without bound or panic on length overflow.
- Bumps h2 0.4.14 to the patched 0.4.16 in `rust/Cargo.lock`. That is the copy actually built (hyper 1.x via reqwest and wiremock).
- Adds `RUSTSEC-2026-0258` to `rust/.cargo/audit.toml` for the second lockfile entry, h2 0.3.27. The 0.3.x line has no patched release (the advisory lists `>= 0.4.16` only), and that entry is an inactive optional dependency of `aws-smithy-http-client` behind its `hyper-014` feature. `cargo tree -i h2@0.3.27` finds no match, so it is never compiled, but cargo-audit reads the lockfile and cannot see feature activation.

## Test Plan
- `cargo audit` in `rust/` exits 0.
- Before the ignore, `cargo audit` reports only h2 0.3.27, confirming the 0.4.x bump resolved the built copy.
- `cargo metadata --locked` succeeds, so the lockfile is consistent.
- `cargo check --locked --all-targets` succeeds.

## Follow-up
- Drop the `RUSTSEC-2026-0258` ignore once `aws-smithy-http-client` stops carrying the optional h2 0.3 dependency.

### DIFF
--- a/rust/.cargo/audit.toml
+++ b/rust/.cargo/audit.toml
@@ -12,4 +12,8 @@ ignore = [
     "RUSTSEC-2026-0098",
     # rustls-webpki 0.101.x is pinned by the transitive rustls 0.21 TLS stack.
     "RUSTSEC-2026-0099",
+    # h2 0.3.x has no patched release and is an inactive optional dependency of
+    # aws-smithy-http-client; the hyper-014 feature that would build it is never enabled.
+    # h2 0.4.x is on the patched 0.4.16.
+    "RUSTSEC-2026-0258",
 ]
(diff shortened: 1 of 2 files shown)


dev-jodee:
done in #246

PR #248, opened 2026-08-18
fix(memory): reject a CryptoKeyPair whose public key does not match its private key

dev-jodee:
## Problem

The memory signer's `keyPair` config source passed the pair straight to `createSignerFromKeyPair`, which derives `address` from `publicKey` and signs with `privateKey` without checking they correspond. The other three sources (raw bytes, base58/U8Array string, keypair file) are validated upstream - `createKeyPairFromBytes` proves the seed matches the public key - so a hand-assembled `{ publicKey: A, privateKey: B }` was the one input of four that could advertise address A while producing signatures only B can verify. The upstream signer closures also capture the caller's pair object, so replacing its `privateKey` after construction re-pointed signing while `address` stayed stale (`Object.freeze` protects the wrapper, not the pair).

Impact is limited: the failure is loud (a B-signature in A's slot never passes validator verification), no generator produces mismatched pairs, and the other language implementations derive the public key from the private key so they cannot express a mismatch. This closes a consistency gap, not an exploit.

## Fix

In the `keyPair` branch, prove the match with a probe: sign 32 random bytes with `privateKey` and verify with `publicKey` - the same guard the byte-based path gets upstream. Mismatches reject with `SIGNER_INVALID_PRIVATE_KEY`; broken/non-Ed25519 pairs keep rejecting with `SIGNER_CONFIG_ERROR` as before. The validated keys are passed on as a fresh pair object, so mutating the caller's `CryptoKeyPair` afterwards cannot affect the signer.

A probe signature is used deliberately instead of exporting and comparing key bytes: `exportKey` throws `CANNOT_EXPORT_NON_EXTRACTABLE_KEY` on non-extractable private keys, which are the `generateKeyPair()` default and the package's documented usage.

## Testing

- New: mismatched pair rejected with `SIGNER_INVALID_PRIVATE_KEY`; mutating the caller's pair after construction leaves the signer signing under its original address (verified signature).
- Package tests 28 passed, typecheck, eslint, prettier, build, umbrella treeshake all green.

### DIFF
--- a/typescript/packages/memory/src/memory-signer.ts
+++ b/typescript/packages/memory/src/memory-signer.ts
@@ -130,28 +130,17 @@ async function resolveKeyPairSigner(config: MemorySignerConfig): Promise<KeyPair
  * `CryptoKeyPair` afterwards cannot re-point the signer's key.
  */
 async function keyPairSignerFromCryptoKeyPair(keyPair: CryptoKeyPair): Promise<KeyPairSigner> {
-    const { privateKey, publicKey } = keyPair;
-
-    let matches: boolean;
+    const probe = new Uint8Array(32); // fixed: the correspondence check needs no entropy
     try {
-        const probe = crypto.getRandomValues(new Uint8Array(32));
-        const probeSignature = await signBytes(privateKey, probe);
-        matches = await verifySignature(publicKey, probeSignature, probe);
-    } catch (error) {
-        throwSignerError(SignerErrorCode.CONFIG_ERROR, {
-            cause: error,
-            message: 'Failed to create signer from keyPair — keyPair must be a valid Ed25519 CryptoKeyPair',
-        });
-    }
-    if (!matches) {
-        throwSignerError(SignerErrorCode.INVALID_PRIVATE_KEY, {
-            message: 'keyPair public key does not match its private key',
-        });
-    }
-
-    try {
+        const { privateKey, publicKey } = keyPair;
+        if (!(await verifySignature(publicKey, await signBytes(privateKey, probe), probe))) {
+            throwSignerError(SignerErrorCode.INVALID_PRIVATE_KEY, {
+                message: 'keyPair public key does not match its private key',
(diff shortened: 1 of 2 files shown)


typescript/packages/memory/src/memory-signer.ts
```diff
@@ -128,6 +122,43 @@ async function resolveKeyPairSigner(config: MemorySignerConfig): Promise<KeyPair
...
 
+/**
+ * The other config sources validate that the public key matches the private key
+ * upstream, but `createSignerFromKeyPair` derives the address from `publicKey`
+ * and signs with `privateKey` without checking they correspond. Prove the match
+ * with a probe signature, and pass a fresh pair object so mutating the caller's
+ * `CryptoKeyPair` afterwards cannot re-point the signer's key.
+ */
+async function keyPairSignerFromCryptoKeyPair(keyPair: CryptoKeyPair): Promise<KeyPairSigner> {
+    const { privateKey, publicKey } = keyPair;
```
amilz: can this go into the try/catch? it hink that'll preserve the current config error we have and then add a test asserting `null` and `{}` still reject with `SIGNER_CONFIG_ERROR`

amilz (commented):
one nit

typescript/packages/memory/src/memory-signer.ts
dev-jodee: Done in 0d8b8c6 — moved the destructure into a single try/catch with a `SignerError` rethrow (same pattern as `resolveKeyPairSigner`), so `null` and `{}` reject with `SIGNER_CONFIG_ERROR` again; added the two tests. Also took the fixed all-zero probe from the suggested shape since the correspondence check needs no entropy. Skipped the `exportKey`/`importKey` round-trip from that suggestion: `verifySignature` accepts the pair's own `publicKey` directly, and re-importing it buys nothing here.

amilz (approved):

PR #252, opened 2026-08-18
fix(fireblocks): select the vault address by asset instead of taking the first

dev-jodee:
## Problem

All four implementations resolved the signer's identity as `addresses[0]` from `addresses_paginated`, with no validation, and modelled each entry as `{ address }` only - `assetId`, `type`, `tag` and `paging.after` were discarded, so an ambiguous or truncated response was undetectable.

The realistic trigger is misconfiguration rather than a hostile provider: `vaultAccountId` is a short numeric string (`"0"` vs `"10"`), and `SOL` vs `SOL_TEST` differs by a suffix. Either mistake produced a **working** signer bound to an unintended identity, silently - and that identity is used as the fee payer. Fireblocks is the only provider-resolved backend here that indexes into a list; Privy, Utila, Crossmint and Dfns all dereference a uniquely-identified wallet.

Nothing is forgeable: the returned signature is always verified against the resolved key. The reachable harm is identity substitution, so severity is low - but it fails silently today, and the precedent for fixing it is already in the repo (Fordefi verifies vault ownership during `init()`).

## Fix

Filter candidate addresses by the configured asset id and accept only a **single distinct** address. An empty or ambiguous result fails with `INVALID_PUBLIC_KEY` naming the vault account and asset, instead of picking an entry.

Two deliberate leniencies:
- Entries **without** an `assetId` are kept. The endpoint is already scoped by asset in its path, and older/partial responses may omit the field; rejecting them would break working setups for no security gain.
- Duplicate entries for the **same** address are not ambiguous (Fireblocks can return multiple entries per address, e.g. per tag).

`assetId` is now modelled in the response types in all four languages. **No configuration or public API changes** - this is internal validation only.

## Not done here (deliberately)

The finding also asked for a required expected-pubkey config field and revalidation before every signature. A *required* field breaks the public API in all four languages for every existing user, and re-fetching the address on each sign adds a network round-trip to defend against mid-session rotation. An *optional* `expectedPublicKey` pin is a reasonable follow-up if wanted; it is not in this PR.

## Testing

Four new tests per language: correct address selected when a foreign-asset entry comes first; ambiguous (two distinct addresses for the configured asset) rejected; no entry for the configured asset rejected; duplicate entries for one address accepted.

Rust 319 tests x sdk-v2/v3/v4 + clippy `-D warnings` + fmt; Python 457 + ruff + mypy strict; Go `-race` + vet + golangci-lint across all modules; TypeScript 44 + typecheck + eslint + prettier + build.

### DIFF
--- a/typescript/packages/fireblocks/src/__tests__/fireblocks-signer.test.ts
+++ b/typescript/packages/fireblocks/src/__tests__/fireblocks-signer.test.ts
@@ -353,6 +353,30 @@ describe('FireblocksSigner', () => {
             expect(signer.address).toBe(wanted.address);
         });
 
+        it('selects the address for a custom assetId, not the default', async () => {
+            const devnet = await generateKeyPairSigner();
+            const mainnet = await generateKeyPairSigner();
+            mockFetch.mockResolvedValueOnce({
+                ok: true,
+                json: async () => ({
+                    addresses: [
+                        { address: mainnet.address, assetId: 'SOL' },
+                        { address: devnet.address, assetId: 'SOL_TEST' },
+                    ],
+                }),
+            });
+
+            const signer = new FireblocksSigner({
+                apiKey: TEST_API_KEY,
+                assetId: 'SOL_TEST',
+                privateKeyPem: TEST_RSA_PRIVATE_KEY,
+                vaultAccountId: TEST_VAULT_ACCOUNT_ID,
+            });
+            await signer.init();
+
+            expect(signer.address).toBe(devnet.address);
+        });
+
         it('rejects an ambiguous address response', async () => {
             const first = await generateKeyPairSigner();
             const second = await generateKeyPairSigner();
--- a/typescript/packages/fireblocks/src/__tests__/fireblocks-signer.test.ts
+++ b/typescript/packages/fireblocks/src/__tests__/fireblocks-signer.test.ts
@@ -326,8 +326,97 @@ describe('FireblocksSigner', () => {
 
             await expect(signer.init()).rejects.toMatchObject({
                 code: 'SIGNER_INVALID_PUBLIC_KEY',
-                message: expect.stringContaining('No addresses found in Fireblocks vault'),
+                message: expect.stringContaining('returned no address'),
             });
+        });
+
+        it('selects the address for the configured asset', async () => {
+            const wanted = await generateKeyPairSigner();
+            const other = await generateKeyPairSigner();
+            mockFetch.mockResolvedValueOnce({
+                ok: true,
+                json: async () => ({
+                    addresses: [
+                        { address: other.address, assetId: 'SOL_TEST' },
+                        { address: wanted.address, assetId: 'SOL' },
+                    ],
+                }),
+            });
+
+            const signer = new FireblocksSigner({
+                apiKey: TEST_API_KEY,
+                privateKeyPem: TEST_RSA_PRIVATE_KEY,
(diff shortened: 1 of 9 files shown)


typescript/packages/fireblocks/src/fireblocks-signer.ts
```diff
@@ -179,16 +180,11 @@ export class FireblocksSigner<TAddress extends string = string> implements Solan
         const uri = `/v1/vault/accounts/${encodeURIComponent(this.vaultAccountId)}/${encodeURIComponent(this.assetId)}/addresses_paginated`;
         const addressesResponse = await this.request<VaultAddressesResponse>('GET', uri);
 
-        const firstAddress = addressesResponse.addresses?.[0]?.address;
-        if (!firstAddress) {
-            throwSignerError(SignerErrorCode.INVALID_PUBLIC_KEY, {
-                message: 'No addresses found in Fireblocks vault',
-            });
-        }
+        const address = this.selectVaultAddress(addressesResponse.addresses ?? []);
```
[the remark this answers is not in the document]
dev-jodee: Not reachable for this signer, so leaving as-is.

Fireblocks refuses to create additional deposit addresses for account-based assets, which is what every Solana asset id is (`SOL`, `SOL_TEST`, SPL). From the [Create new asset deposit address](https://developers.fireblocks.com/api-reference/vaults/create-new-asset-deposit-address) reference:

> Should be used for UTXO or Tag/Memo based assets ONLY. Requests with account based assets will fail.

So a Solana vault account holds exactly one address for its asset, the response is a single entry, and there is no later page to miss. If someone points `assetId` at a genuinely multi-address asset (a UTXO coin), its addresses are not base58 32-byte Solana pubkeys, so `assertIsAddress` (and the equivalent parse in the other three languages) rejects them regardless of paging.

Fair read of the generic endpoint contract though - the restriction lives in the sibling endpoint's docs, not this one's.

typescript/packages/fireblocks/src/__tests__/fireblocks-signer.test.ts
amilz, line 1: lgtm--though all of these use the default asset ID `SOL`--you think it's worth duplicating these tests w/ custom config `assetId`?

amilz (dismissed):

typescript/packages/fireblocks/src/__tests__/fireblocks-signer.test.ts
dev-jodee: Good catch, and it turned out to be a real hole rather than just duplication. Added in d67b7d5.

I checked whether the existing tests could tell the difference by replacing the configured asset comparison with a hardcoded `"SOL"` literal in all four languages: **every test still passed** (TS 44, Python 29, Rust 21, Go). Because every test configured the default `SOL`, the value under test happened to equal what the hardcoded version returns, so the config plumbing was never actually exercised end to end.

One test per language now configures `SOL_TEST` and serves both a `SOL` and a `SOL_TEST` entry, asserting the `SOL_TEST` address wins. Re-ran the same mutation afterwards and all four suites now fail, which is the property that was missing. It also gives `SOL_TEST` (the documented devnet asset id) its first coverage.

I did not duplicate the ambiguous/empty/duplicate cases under a custom asset id: those branches are already covered and re-running them with a different asset adds no new coverage, whereas the select path was the one that could hide a hardcoded default.

amilz (approved):

PR #254, opened 2026-08-18
feat(crossmint): report unconfirmed broadcasts with the provider transaction id

dev-jodee:
Stacked on #245 (and through it on #241/#231). Applies the #241 treatment to Crossmint, the other broadcast-managed backend, as promised there.

## Problem

Crossmint executes transactions server-side. Once a create is accepted, any later failure - poll timeout, terminal `failed` status, awaiting-approval with no key of ours, extraction or verification failure - surfaced as a generic `SIGNING_FAILED`/`REMOTE_API_ERROR` with the Crossmint transaction id unreachable. A caller cannot distinguish "never created" from "created, outcome unknown", and the natural blind retry creates a second transaction Crossmint may also execute: a duplicate spend.

## Fix

Two layers, all four languages:

1. **`BROADCAST_UNCONFIRMED` after create.** Every post-create failure now maps to the existing `BROADCAST_UNCONFIRMED` error carrying the Crossmint transaction id (Rust `BroadcastUnconfirmed { provider_tx_id }`, Python `provider_transaction_id`, Go `ProviderTxID`, TS `context.providerTransactionId`), with the original error preserved as detail/cause. Create-time failures are untouched: nothing was created, retry is safe.
2. **`x-idempotency-key` on create.** Crossmint's create-transaction endpoint documents this header ("Unique key to prevent duplicate transaction creation"). Each create now sends a deterministic key derived from the message bytes (UUID from the first 16 bytes of SHA-256), so a blind retry of the exact same bytes reuses the key and Crossmint deduplicates the create. A retry built with a fresh blockhash is a new transaction, which is what `BROADCAST_UNCONFIRMED` protects against.

Docs on the signing entry points and the package README state the retry contract. The key derivation is a shared core helper in each language (Fordefi's native path uses it too), since this PR would otherwise duplicate it.

## Testing

Per language: post-create failures (timeout, failed status, awaiting-approval, unverifiable artifacts) assert `BROADCAST_UNCONFIRMED` with the transaction id reachable; the create request asserts the exact derived idempotency key (derivation mirrored in the test, not calling the source helper).

Full local gates: TS build/test/typecheck/eslint/prettier (workspace), Rust crossmint suite on sdk-v2/v3/v4 + fmt, Python 473 tests + ruff + mypy strict, Go test -race + vet + gofmt.


amilz (approved):

dev-jodee:
Superseded by #269, which carries this branch's commits unchanged (this branch is an ancestor of #269's tip). Closing to land the stack as a single squash into `main`. Branch kept for now.

PR #256, opened 2026-08-19
feat(signers): expose broadcast transaction capability

dev-jodee:
## Summary
- add a default broadcast capability marker to the Rust trait and Python ABC
- report Crossmint and native-mode Fordefi as broadcast-managed signers
- delegate the capability through the unified Rust Signer and add regression coverage

## Test Plan
- just rust-test
- just rust-fmt
- just py-test
- just py-fmt


dev-jodee:
Replaced by a gh-stack PR stacked above #255.

PR #257, opened 2026-08-19
feat(signers): expose broadcast transaction capability

dev-jodee:
## Summary
- add a default broadcast capability marker to the Rust trait and Python ABC
- report Crossmint and native-mode Fordefi as broadcast-managed signers
- delegate the capability through the unified Rust Signer and add regression coverage

## Test Plan
- just rust-test
- just rust-fmt
- just py-test
- just py-fmt

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #258, opened 2026-08-19
feat(signers)!: reject versioned wire transactions from providers

dev-jodee:
## Summary

Solana's v1 transaction format ([SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md)) inverts the envelope: signatures move to the tail and `0x80 | version` lands at offset zero. Every backend here that round-trips a full serialized transaction through a provider API decodes the response with a reader that assumes the legacy layout (compact-u16 signature count, signatures, message).

Nothing detected the version prefix, so a v1 response surfaced as a misleading failure rather than an accurate one:

- TypeScript Fordefi's hand-rolled parser read `0x81` as the start of a compact-u16 count, consumed the header byte as a continuation, and reported "truncated transaction" (or, for other header values, returned bytes from the wrong offset).
- Rust Fordefi reported "versioned/v0 transactions are not supported", blaming v0 for a v1 envelope.
- The remaining backends produced an opaque deserialization error.

This adds a shared guard per language that rejects a version-prefixed envelope before decoding, naming both the provider and the detected version. Legacy and v0 envelopes are unaffected: they cap at 12 signatures, so the high bit of the first byte is never set, which is the same offset-zero discriminator the SIMD intends for infrastructure.

Wired into every provider-response decode site:

| Language | Sites |
| --- | --- |
| Rust | `TransactionUtil::reject_versioned_wire_transaction` in Privy, Turnkey, CDP, Utila, Fordefi, Crossmint (both paths) |
| TypeScript | `assertUnversionedWireTransaction`, called inside the core signature extractors (covers Privy, Turnkey, CDP, Utila, Fordefi) plus the Crossmint and Fordefi decoders |
| Python | `assert_unversioned_wire_transaction` in the same six backends |
| Go | `core.AssertUnversionedWireTransaction` in the same six backends |

Also corrects two now-inaccurate claims that this change contradicts: the Rust Fordefi comment describing a v0 wire transaction as version-prefixed at offset zero (only v1+ is), and the Go Utila doc claiming solana-go decodes legacy and all versioned messages alike.

This is the fail-loud slice only. It does not add v1 support: that needs an SDK upgrade (Agave 4.2.x in Rust, a kit floor raise in TypeScript, solders' `MessageV1` in Python, a solana-go library decision), a versioned transaction type on the public signing API, and per-provider confirmation that each vendor accepts v1.

## Breaking change

The TypeScript core helpers `extractSignatureFromTransactionBytes` and `extractSignatureFromWireTransaction` now require a `providerName` field, so the rejection error names the provider like every other error in the library.

## Test Plan

- `just rust-test` (all three SDK matrices: 335 passed each)
- `just rust-fmt`
- `just ts-test`, `just ts-fmt`, `just ts-treeshake`
- `just py-test` (480 passed), `just py-fmt`
- `just go-test`, `just go-fmt`

New tests per language cover: a real legacy transaction is accepted (no false positive), a `0x81` envelope is rejected with a serialization error naming provider and version, and empty bytes defer to the decoder.

---

### DIFF
--- a/go/core/txutil.go
+++ b/go/core/txutil.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"encoding/base64"
+	"fmt"
 
 	"github.com/gagliardetto/solana-go"
 )
@@ -15,6 +16,21 @@ func Serialize(tx *solana.Transaction) (string, error) {
 		return "", WrapSignerError(CodeSerializationError, "failed to serialize transaction", err)
 	}
 	return base64.StdEncoding.EncodeToString(b), nil
+}
+
+// AssertUnversionedWireTransaction rejects a wire transaction whose envelope carries
+// a version prefix. Legacy and v0 envelopes both open with a compact-u16 signature
+// count, capped at 12 signatures, so the high bit of the first byte is never set. v1
+// moves its signatures to the tail and puts 0x80|version at offset zero, a layout the
+// signature-slot readers here cannot interpret.
+func AssertUnversionedWireTransaction(provider string, wireBytes []byte) error {
+	if len(wireBytes) == 0 || wireBytes[0]&0x80 == 0 {
+		return nil
+	}
+	return NewSignerError(CodeSerializationError, fmt.Sprintf(
+		"%s returned a v%d transaction envelope, which is not supported yet (only legacy and v0 transactions can be verified)",
+		provider, wireBytes[0]&0x7f,
+	))
 }
 
 // SigningPosition returns the index in the transaction's required-signer list where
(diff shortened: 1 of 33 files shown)


dev-jodee:
Closing: this took the wrong approach. It rejects v1 envelopes, where the goal is to support them. Superseded by the v1 support work stacked on #257. The accurate-error fixes here (Rust Fordefi misattributing v1 to v0, the stale Go Utila doc comment) carry into that work.

PR #269, opened 2026-08-22
feat(signers)!: broadcast-managed signer capability and v1 transaction support

dev-jodee:
## What this is

A single PR carrying the whole broadcast-capability and v1-transaction workstream: 29 commits, 196 files, all four language implementations. It supersedes the stack #231 → #241 → #245 → #254 → #255 → #257, whose branches were all ancestors of this one. Landing it as one PR avoids six sequential approval rounds against `main`, and works around the fact that `/typescript/` in CODEOWNERS is owned solely by @amilz, who authored #245.

There are two independent themes plus a few fixes.

## 1. Broadcast-managed signers

Some providers do not just sign: they submit. Crossmint always does, and Fordefi does in native mode. Before this change the interfaces could not express that, so a failure after the provider had accepted the transaction was indistinguishable from a failure before, and callers could retry into a duplicate spend.

**Capability marker.** Signers now declare whether signing also broadcasts.

- Rust: `SolanaSigner::broadcasts_transactions() -> bool`, defaulting to `false`, delegated through the unified `Signer` enum.
- Python: `SolanaSigner.broadcasts_transactions` on the ABC, same default.
- Go: a separate `core.TransactionBroadcaster` interface with `BroadcastsTransactions() bool`, so existing signers are unaffected.
- TypeScript: a new `SolanaSendingSigner<TAddress>` in `@solana/keychain-core` extending `TransactionSendingSigner`, deliberately *not* extending `MessagePartialSigner`, plus an `isSolanaSendingSigner()` guard.

**Unconfirmed-broadcast error.** A new terminal error state for "the provider may have executed this; do not blindly retry".

- Rust: `SignerError::BroadcastUnconfirmed { provider_tx_id, provider_status, detail }`. `Debug` prints the provider transaction id and `[REDACTED]` for the detail, so no key material or raw remote response can leak.
- Python: `SignerErrorCode.BROADCAST_UNCONFIRMED` plus `provider_may_have_accepted(status_code)`.
- Go: `core.CodeBroadcastUnconfirmed`, `NewBroadcastUnconfirmedError`, and `UnconfirmedUnlessRejected(status, err)`.
- TypeScript: `BROADCAST_UNCONFIRMED` code plus `providerStatus()` and `providerMayHaveAccepted()` helpers.

Both id fields are `None`/empty when the create failed with no usable response at all: there is nothing to reconcile against, and the only safe recovery is replaying the identical bytes.

**Behaviour fixes that fall out of this.**

- Crossmint managed-broadcast signing actually works now, returns the landed transaction's fee-payer signature, and narrows returned message versions before serialization.
- Crossmint keeps scanning submitted approvals past an incomplete entry instead of stopping at the first one.
- Fordefi sends a deterministic idempotence id derived from the message bytes on native creates, leaves the caller's transaction untouched in native mode, classifies native results as complete without the caller's slots, and carries the provider transaction id through asyncio cancellation.
- The Go batch helper rejects broadcast-managed signers outright, since batching a broadcasting signer has no safe semantics.
- Idempotency key derivation is shared in core rather than duplicated per signer.

## 2. v1 transaction support (SIMD-0385)

`sign_transaction` now takes a versioned transaction everywhere, accepting legacy, v0 and v1.

- Rust: `&mut Transaction` becomes `&mut VersionedTransaction`. Convert a legacy transaction with `.into()`. **v1 requires the `sdk-v4` feature**; `sdk-v2` and `sdk-v3` still build and still handle legacy and v0.
- Python: `sign_transaction(transaction: VersionedTransaction)`, with `transaction_util` moved onto `VersionedMessage` and `to_bytes_versioned` so the version prefix comes from the message's own serialization rather than a hardcoded `0x80`. Requires `solders>=0.29`. CDP signed transactions are decoded with the versioned parser.
- Go: v1 support via the unreleased `solana-go`, reusing the shared signer-position helper.
- TypeScript: v1 parsing across the signers, with Fordefi taking the broadcast id from kit instead of indexing the account map.

## 3. Other

- Fireblocks supports `PROGRAM_CALL` signing in sign-only mode (Rust, TypeScript, Python, Go).
- `@solana/kit` moves off the canary pin to 8.0.0 across every package.
- Node 18 abort composition fixed, and honest Fordefi factory types.

## Breaking changes

- `sign_transaction` takes a versioned transaction in Rust, Python and Go. Legacy callers need a conversion at the call site.
- Rust v1 support is gated behind `sdk-v4`.
- Python requires `solders>=0.29`.
- TypeScript managed-broadcast signers move to `SolanaSendingSigner`, which no longer implies `MessagePartialSigner`.
- The Go batch helper now returns an error for broadcast-managed signers.

## Attribution

The `SolanaSendingSigner` core type and the TypeScript managed-broadcast modelling are @amilz's work from #245, preserved as their original commits in this branch.

## Test plan

- `just test` (rust + ts unit)
- `just rust-test` across the `sdk-v2`, `sdk-v3` and `sdk-v4` matrices
- `just ts-test` and `just ts-treeshake`
- `just py-test`
- `just fmt`
- Full CI was green on this exact tree as #257 (152/152 checks).


amilz (approved):
Broadcast-managed signer capability + v1 (SIMD-0385) transaction support across Rust/Python/Go/TS; all 152 CI checks green, breaking changes documented, error-redaction covered by tests. LGTM 🔥

PR #276, opened 2026-08-24
feat(signers): sign-and-send parity for Rust, Python and Go

dev-jodee:
## Summary
- Ports the `signAndSendTransaction` helper from #270 to the Rust, Python and Go cores: `sign_and_send`, `sign_and_send_transaction`, `core.SignAndSendTransaction`. Each uses the broadcast capability marker the packages already carry (`broadcasts_transactions()`, `broadcasts_transactions`, `TransactionBroadcaster`), so no duck-typed capability helper is needed. No package gains an RPC dependency: the broadcast is a caller-injected function.
- Fixes a Go cancellation bug found while reviewing the same PR: cancelling the context mid-poll on a signer that broadcasts server-side returned `SIGNER_HTTP_ERROR`, so cancellation was indistinguishable from a request that never landed and the caller was free to retry a transaction that may already be executing. Crossmint and Fordefi native mode now return `SIGNER_BROADCAST_UNCONFIRMED` with the provider transaction id.
- The other two parts of #270 are deliberately not ported: abort support is already native in each language (`context.Context`, dropping the future, `asyncio` cancellation), and the deprecated class tier being removed there never existed in these packages.

## Test Plan
- `just rust-test` (sdk-v2/v3/v4 matrices), `just rust-fmt` clean
- `just py-test` (510 passed), `just py-fmt` clean (ruff + mypy strict)
- `go build ./...` and `go test ./...` across every Go module, `gofmt`/`go vet` clean
- New unit tests cover each branch: provider broadcast used as-is, injected sender receives the encoded transaction, missing sender rejected before signing, partial signature never broadcast, and cancelled broadcast polling reported as unconfirmed with the provider transaction id

## Breaking Changes
- None. All additions, plus one error code change on a cancellation path that previously reported a retry-safe failure.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -194,10 +194,10 @@ present.
 
 ### Sign and Send
 
-`core.SignAndSendTransaction` gets a transaction on chain with one call whichever
-shape the signer has. Managed-broadcast signers (Crossmint, Fordefi native mode)
-broadcast through their provider and ignore the send function; every other signer
-signs and the injected function broadcasts the result:
+`core.SignAndSendTransaction` gets a transaction on chain with one call.
+Managed-broadcast signers (Crossmint, Fordefi native mode) broadcast through their
+provider and ignore the send function; every other signer signs and the send
+function broadcasts the base64-encoded result:
 
 ```go
 sig, err := core.SignAndSendTransaction(ctx, signer, tx,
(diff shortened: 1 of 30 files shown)


dev-jodee:
@greptile re-review

amilz (approved):

PR #279, opened 2026-08-25
feat(fordefi)!: native manual push mode for Rust

dev-jodee:
Rebase of #271 by @hvbr1s. **Credit for the feature is entirely @hvbr1s'**; the original commit authorship is preserved and #271 carries the review history.

This PR is now scoped to **Rust only** and is stacked on #287 (`feat/signer-trait-split`), which split `SolanaSigner` into a base trait plus one capability trait per backend. Manual mode is the feature that motivated the third capability, so it lands on top of the split rather than against the old single-trait `FordefiSigner`. TypeScript, Python and Go ports land as sibling PRs stacked on their own language split PRs.

## What this adds

Fordefi's `push_mode: "manual"` path alongside the existing black box and native auto modes, as a third Fordefi signer type: `FordefiNativeManualSigner`, the first implementor of `ModifyingSigner`.

In manual mode Fordefi may replace the recent blockhash and manage the Compute Budget fee instructions, then signs **without** broadcasting. `modify_and_sign_transaction` replaces the caller's `&mut VersionedTransaction` with the transaction Fordefi signed and returns the serialized bytes for the caller to broadcast. There is no `sign_transaction` and no `sign_and_send_transaction` on this type: absent capability, not an erroring stub.

## Contract

Fordefi is a trusted provider here, consistent with the rest of the codebase: **the rewrite is not diffed and not validated**. There is no field-by-field comparison and no fee cap. What is enforced:

- **Preconditions before submit.** The configured vault must be the transaction's fee payer (first static account key), and no signature may be present yet: Fordefi may only rewrite a message nobody has signed, and must sign before every downstream signer.
- **Signature binding after.** The returned signature is located by the vault's required-signer position and verified with ed25519 against the **returned** message bytes. A signature that does not cover what came back is rejected, and a rejected result leaves the caller's transaction untouched.
- **Replacement, not merge.** On success the caller's transaction becomes the one the signature covers. The signature is never merged into the caller's original.
- **Namespaced idempotency.** The manual create's `x-idempotence-id` is namespaced by chain and vault, so the same message bytes submitted for signing cannot collide with an earlier auto create that did broadcast them.

## Modes

| Mode | Type | Capability trait | Entry point |
| --- | --- | --- | --- |
| Black box (`chain` unset) | `FordefiBlackBoxSigner` | `TransactionSigner` | `sign_transaction` |
| Native auto (`chain` set, `push_mode` `Auto`/`None`) | `FordefiNativeAutoSigner` | `SendingSigner` | `sign_and_send_transaction` |
| Native manual (`chain` set, `push_mode` `Manual`) | `FordefiNativeManualSigner` | `ModifyingSigner` | `modify_and_sign_transaction` |

`Signer::from_fordefi` dispatches on `chain` and `push_mode`; each concrete constructor rejects the other modes' configs with `ConfigError`.

The umbrella `Signer` enum gains the `FordefiNativeManual` variant and `as_modifying_signer() -> Option<&dyn ModifyingSigner>` beside the existing accessors. Its inherent `sign_and_send` now routes sending, then transaction, then modifying, with the modifying path running `modify_and_sign_transaction` and handing the rewritten encoding to the caller-supplied send closure.

## Verification

- 355 passed `sdk-v2`, 355 passed `sdk-v3`, 360 passed `sdk-v4` (0 failed, 0 skipped), plus doc-tests.
- `just rust-fmt`: `cargo fmt` clean, `cargo clippy --all-targets -D warnings` clean on all three SDK matrices with `all,unsafe-debug,integration-tests`.
- Six new unit tests cover the replacement guarantee, the namespaced idempotency key, both preconditions, rejection of a signature over other bytes, and umbrella routing through `as_modifying_signer()`.
- One devnet integration test (`test_fordefi_manual_sign_and_broadcast`, marker-gated) signs manually, broadcasts the returned transaction and confirms it.

Docs updated: `CLAUDE.md`, `docs/ADDING_SIGNERS.md`, `docs/SECURITY_MODEL.md`, `rust/README.md`.


typescript/packages/fordefi/src/index.ts
```diff
@@ -1,5 +1,11 @@
-export { createFordefiSigner } from './fordefi-signer.js';
-export type { FordefiNativeSigner, FordefiRequestSigner, FordefiSignerConfig } from './fordefi-signer.js';
+export { DEFAULT_MAX_PRIORITY_FEE_LAMPORTS, createFordefiSigner } from './fordefi-signer.js';
+export type {
+    FordefiManualSignerConfig,
+    FordefiNativeManualSigner,
+    FordefiNativeSigner,
+    FordefiRequestSigner,
```
amilz, line 6: note te self--come back to...i'm not conviced we want to export these.

amilz (approved):

PR #284, opened 2026-08-26
refactor(crossmint)!: make Crossmint a sending-only signer

dev-jodee:
## Problem

Crossmint's Wallets API has no sign-only path. Creating a transaction and approving it always causes Crossmint to execute it server-side ([create-transaction](https://docs.crossmint.com/api-reference/wallets/create-transaction): "Transaction will be automatically broadcast once it has all necessary approvals"); there is no `signOnly` parameter, and Solana wallets are routed through `/transactions` rather than the EVM-only `/signatures` endpoint.

Despite that, `sign_transaction` submitted, polled and returned. A caller asking to sign got a transaction that had already landed, signalled only by an empty encoded transaction. Two concrete consequences:

- Callers could not distinguish "signed, yours to broadcast" from "already on chain", and re-sending was a duplicate.
- In the `onChain.txId` fallback branch the signature was placed into the caller's transaction even though it identifies the landed transaction and does not necessarily cover the caller's message.

## Change

Crossmint becomes sending-only in Rust, Python and Go:

- `sign_transaction` / `SignTransaction` rejects with a signing failure, matching what Fordefi native mode already did.
- `sign_and_send_transaction` / `SignAndSendTransaction` owns the managed flow and returns the signature identifying the landed transaction.
- The caller's transaction is never mutated, and the empty-encoded-transaction convention is gone.
- `broadcasts_transactions()` / `BroadcastsTransactions()` stays `true`, so the existing send helpers and batch-signing guards dispatch unchanged.
- Rewritten-transaction handling, `BROADCAST_UNCONFIRMED` reconciliation and idempotency-key behavior are unchanged.

TypeScript needed no change: `CrossmintSigner` already implements `SolanaSendingSigner` and exposes no `signTransactions`. This brings the other three languages to that shape.

No trait or interface split: the existing capability flag already discriminates, and splitting would break Rust's `Signer` enum, Go's `core.Signer` factory return type and the Python ABC without buying enforcement that survives those unified surfaces.

## Tests

- New per-language test asserting `sign_transaction` is rejected and leaves the caller's transaction unsigned.
- Existing managed-flow tests moved to `sign_and_send_transaction`; every success path now also asserts the caller's transaction stays unsigned.
- Integration tests split: broadcast-managed backends exercise the send path, everything else keeps the sign path.

`just rust-test` (sdk-v2/v3/v4), full Python suite (514 passed), all Go modules, `just rust-fmt` and `just py-fmt` clean.

## Breaking change

`sign_transaction` / `SignTransaction` on the Crossmint signer now fails. Callers move to `sign_and_send_transaction` / `SignAndSendTransaction`, which returns a bare signature rather than a signed-transaction shape and leaves the caller's transaction unsigned.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -197,14 +197,16 @@ at runtime; the second is fixed per backend:
 |---------|----------------------------|-------------------|--------------------------|---------------|
 | memory, vault, privy, turnkey, awskms, fireblocks, gcpkms, dfns, para, openfort | not implemented | yes | not implemented | yes |
 | cdp | not implemented | yes | not implemented | UTF-8 payloads only, otherwise `CodeSerializationError` |
-| crossmint | `true` | yes | yes | `CodeSigningFailed` |
+| crossmint | `true` | `CodeSigningFailed` | yes | `CodeSigningFailed` |
 | utila | not implemented | yes | not implemented | `CodeSigningFailed` |
 | fordefi (black-box mode) | `false` | yes | `CodeSigningFailed` | yes |
 | fordefi (native mode) | `true` | `CodeSigningFailed` | yes | yes |
 
-Crossmint supports both: it decides per request whether to rewrite and broadcast
-the transaction or to sign the caller's exact bytes, and `SignTransaction`
-exposes that distinction through an empty `EncodedTransaction`.
+Crossmint executes every approved transaction server-side and exposes no
+sign-only API, so it offers `SignAndSendTransaction` only. It may rewrite the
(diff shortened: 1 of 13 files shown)


amilz (approved):
Coherent breaking refactor: Crossmint becomes sending-only across Rust/Go/Python (sign_transaction rejects, sign_and_send owns the managed flow, caller tx never mutated). Well-tested, docs updated, all CI + Greptile green.

PR #310, opened 2026-09-10
build: drop aws-lc-rs and prune unused dependencies

dev-jodee:
Follow-up to #309, which has merged; this is now rebased directly on `main`.

## Why

The `--timings` run in #309 showed the cold build is a single serial dependency chain with `solana-keychain` at the very end of it. `aws-lc-sys` sits at the head: a C build with no Rust caller of its own, pulled in only because two of our dependencies default to the aws-lc-rs rustls provider. This is also what dominates downstream consumers' builds (kora's cold build is gated by the same chain, entering through this crate).

`aws-lc-rs` had two independent doors, and closing one alone changes nothing:

1. `google-cloud-kms-v1`'s default feature `default-rustls-provider` -> `gax-internal/_default-rustls-provider` -> `google-cloud-auth/default-rustls-provider` = `rustls/aws_lc_rs`. `google-cloud-auth`'s other default, `default-idtoken-backend`, is a literal `dep:aws-lc-rs`.
2. `aws-sdk-kms`'s default `default-https-client` -> `aws-smithy-runtime/default-https-client` -> `aws-smithy-http-client/rustls-aws-lc`.

Neither crate exposes a ring alternative through its own features, so both need `default-features = false` plus explicit wiring.

## Changes

**GCP.** `google-cloud-kms-v1` and `google-cloud-auth` with `default-features = false`. With no provider feature selected, `reqwest` 0.13 falls back to the process-default `CryptoProvider`, so `GcpKmsSigner::from_config` installs ring:

```rust
let _ = rustls::crypto::ring::default_provider().install_default();
```

`install_default` returns `Err` when a provider is already installed and the result is deliberately discarded: an embedding application that installed aws-lc-rs (or anything else) keeps its choice. This is a process-global side effect from a library, which is the honest cost of this change; it is confined to the `gcp_kms` feature, only fires on the `from_config` path, and `with_client` is untouched for callers who build the client themselves.

**AWS.** `aws-sdk-kms` and `aws-config` with `default-features = false` plus the non-TLS parts of their defaults (`rt-tokio`, `credentials-process`, `sso`). `AwsKmsSigner` builds the connector explicitly:

```rust
aws_smithy_http_client::Builder::new()
    .tls_provider(tls::Provider::Rustls(tls::rustls_provider::CryptoMode::Ring))
    .build_https()
```

Dropping the SDK's `rustls` feature also drops `connector-hyper-0-14-x`, which is what was pulling **hyper 0.14 and h2 0.3** alongside hyper 1.x / h2 0.4. That resolves the duplicate-major question in passing rather than being upstream-blocked. The wiremock tests build their client from `aws_sdk_kms::config::Builder` directly, so they were relying on `default-https-client` too and now share the same connector helper.

**cargo-machete.** All hits are false positives of the two known shapes, recorded in `[package.metadata.cargo-machete]` with the reason inline:

| dependency | why it stays |
|---|---|
| `google-cloud-auth` | the handle that turns its own aws-lc-rs defaults off |
| `solana-shred-version-v3`, `solana-signature` | sdk-v3 companion-crate pins (existing comments) |
| `agave-feature-set-pin` | pins the litesvm/agave cluster to 3.0.11 |
| `getrandom` | wasm32-unknown-unknown target dependency enabling the `js` backend |

Nothing was removed on machete's say-so; each was confirmed by build, not grep.

## Numbers

`cargo build --locked --features all,sdk-v2,unsafe-debug` into a fresh target dir, warm registry, 10-core machine, back to back, `/usr/bin/time`:

| | wall | CPU (user+sys) | `target/` | lockfile |
|---|---|---|---|---|
| `main` | 67.7 s | 326.3 s | 2.0 GB | 876 |
| #309 (`debug = "line-tables-only"`) | 65.2 s | 303.3 s | 1.6 GB | 876 |
| this PR | **51.7 s** | **278.4 s** | **1.4 GB** | **858** |

Critical path before (at #309's base):

```
 13.8 ->  56.9 (43.1s) aws-lc-sys 0.43.0
 ...        ->          aws-lc-rs -> rustls-webpki -> rustls
            ->          tokio-rustls -> hyper-rustls -> reqwest 0.13
            ->          google-cloud-auth -> gax-internal -> longrunning -> lro
            ->          google-cloud-kms-v1 -> solana-keychain
```

After:

```
  6.9 ->   8.8 ( 1.9s) cc 1.4.4
 18.6 ->  32.0 (13.5s) ring 0.17.14
 31.9 ->  41.2 ( 9.3s) rustls 0.23.40
 38.1 ->  43.3 ( 5.2s) reqwest 0.13.4
 40.9 ->  53.8 (12.9s) google-cloud-auth 1.13.0
 52.7 ->  59.9 ( 7.2s) google-cloud-kms-v1 1.11.0
 59.7 ->  66.5 ( 6.8s) solana-keychain
```

The chain shape is unchanged; the 43 s head is gone and the rest is smaller. `aws-lc-sys`, `aws-lc-rs`, `hyper 0.14`, `h2 0.3`, `native-tls`, `hyper-tls` and `tokio-native-tls` are all absent from `Cargo.lock`. **Caveat on the wall-clock figure:** the `aws-lc-sys` build script parallelises internally, so on 10 cores it overlaps with the rest of the graph and the 43 s does not show up 1:1 in wall time. On a 2-core GitHub runner it has nothing to overlap with, so the CI saving should be larger than 16 s. That last part is an expectation from the shape of the graph, not something measured here.

## Left alone deliberately

- **`reqwest` 0.12 and 0.13 both remain.** Collapsing onto 0.13 looks tempting (`reqwest` is `pub(crate)`-only here, so it is not an API break) but 0.13 removed the ring feature path: its only rustls options are `rustls` (= aws-lc-rs) and `rustls-no-provider`. Bumping would push the process-global `install_default` from the one `gcp_kms` path onto every remote backend. Not worth it for one deduplicated crate.
- **`openssl` stays in the lockfile**, reached only through `litesvm` -> `agave-precompiles` -> `solana-secp256r1-program` in dev-dependencies. Not in any shipped build.
- **`http` 0.2 stays**, from the AWS SDK core crates. Upstream-blocked.
- No `mockall` in this crate, and no `full` features on tokio or tower: `tokio` is already `default-features = false, features = ["time"]` and neither tower nor tower-http is a direct dependency.
- The set of backends `all` includes is unchanged.

## Testing

- `cargo test --features all,sdk-v2,unsafe-debug`: 385 passed
- `cargo test --no-default-features --features all,sdk-v3,unsafe-debug`: 385 passed
- `cargo test --no-default-features --features all,sdk-v4,unsafe-debug`: 390 passed
- `cargo clippy --all-targets -- -D warnings` for all three SDK matrices with `integration-tests`: clean
- `cargo fmt --all --check`: clean
- `cargo machete --with-metadata`: clean
- Live AWS/GCP KMS integration tests have not been run; the connector and provider changes are exactly the paths the mocked tests cannot cover, so those are worth a run before merge.

## AI disclosure

- [x] AI tooling was used. Tool and extent: Claude Code investigated the `--timings` dependency chain, drafted the feature wiring and this description; I read the full diff, ran the test matrices and machete locally, and can account for every line.


amilz (dismissed):
Build: drops aws-lc-rs in favor of ring for the KMS rustls backends (AWS explicit connector, GCP process-default CryptoProvider) and prunes unused deps; CLAUDE.md security note updated to match. Greptile review green. LGTM.

amilz (approved):
Drops aws-lc-rs and wires ring explicitly on both KMS backends; prunes unused deps. Change is clean; only red check is a v2 fordefi integration flake (v3/v4 fordefi green, PR untouched by fordefi).

PR #314, opened 2026-09-15
fix: improvements & hardening pre-release

dev-jodee:
## What and why

Two independent audit fixes, one commit each.

### 1. Normalize configured base URLs (`855f3a5`)

Only Fordefi and Crossmint trimmed a trailing `/` from a configured `base_url` in Rust. Every other backend left it in place, producing a doubled slash in each request path, and the two that did trim each had their own ad-hoc `trim_end_matches('/')` call.

- Adds a shared `normalize_base_url` helper to the Rust remote-signer utilities and routes every backend through it. Six backends (Fireblocks, Turnkey, Privy, CDP, Dfns, Vault) were not normalizing at all.
- Collapses the five existing ad-hoc trim blocks (Fordefi, Crossmint, Openfort, Para, Utila) onto the same helper, so whitespace handling is consistent too.
- Wires the four TypeScript packages that skipped it (dfns, fireblocks, privy, turnkey) into the existing `normalizeBaseUrl`.
- Trims the Go Vault signer's address directly: `NormalizeHTTPSBaseURL` forces HTTPS, and Vault is routinely reached over plain-HTTP loopback.
- Python already normalized in every backend. No change there.

Behavior change: a base URL carrying a path prefix behind a gateway (e.g. `https://gw.example.com/fireblocks/`) now works where it previously 404'd. Intended, but not a pure no-op.

### 2. Verify the configured public key belongs to the provider key (`6410e12`)

The AWS KMS, GCP KMS, Turnkey, Vault and Fordefi backends all accept a Solana public key from config, never send it to the provider, and never check it against the key the provider actually holds. A mismatch surfaces only when the returned signature fails verification, after a wasted round trip; for `FordefiNativeAutoSigner`, after Fordefi has already broadcast.

Each backend's availability probe now confirms the configured key is the one the provider signs with.

| Backend | How |
|---|---|
| GCP KMS | `GetPublicKey` was already fetched and its `pem` discarded. Now compared. |
| Vault | `/v1/transit/keys/{name}` was already fetched and its `keys` discarded. Now compares the **latest** version, the one signing uses. |
| AWS KMS | New `GetPublicKey` call after `DescribeKey`. |
| Turnkey | New `get_private_key` query after `whoami`. Matches on the `ADDRESS_FORMAT_SOLANA` address or the hex public key. Skipped when `signWith` is already the Solana address. |
| Fordefi | The vault response was already fetched but only `id` was deserialized. Now reads the base58 `address` on chain-specific vaults, the base64 `public_key_compressed` on black box vaults. |

Applied identically in Rust, TypeScript, Python and Go. Shared Ed25519 SPKI decoding lands in each language's core (`signature_util` / `encoding.ts` / `core.sign`).

**Compatibility.** Two backends need a permission they did not need before, and in both cases a caller lacking it sees the health check start returning `false` while signing is unaffected:

- **AWS KMS** — `kms:GetPublicKey` on the key policy.
- **Turnkey** — read access to `get_private_key`. Despite the name, that query returns the private key's *metadata* (public key, curve, derived addresses) and never key material; export is a separate activity targeting an enclave key. But it is a read, so an API key the policy engine scopes to signing activities alone may be denied it.

**Scope.** This is a health check, not a signing guard. Construction still makes no network calls anywhere and the trusted-provider model is unchanged. `is_available()` sits on no signing path, so it only helps callers who already run it. Every returned signature is still ed25519-verified against the configured key before it is attached.

Also removes a stale doc comment in `typescript/packages/fordefi/src/types.ts` claiming `FordefiVaultResponse` was used for "authoritative verification that a configured Solana `publicKey` actually belongs to the vault". No code implemented that until this change.

No change to signature binding in either fix: every affected backend signs its path or body independently of the base URL, so no golden vectors or wire formats move.

## Testing

Run against the combined branch, not the two changes in isolation:

- `just rust-test` — all three SDK matrices pass (402 / 402 / 407, 0 failed).
- `just rust-fmt` — `cargo fmt --check` and clippy clean.
- `just ts-build`, `just ts-test`, `just ts-treeshake` — all packages pass.
- New assertions for both fixes across Rust, TypeScript, Python and Go; each fails without its change.
- Go and Python were verified by CI rather than locally (no Go toolchain, no usable venv on the dev machine). Both were green on the equivalent tree in #315: `go-format`, `go-lint`, all 16 `go-test` jobs, `python-lint`, `python-test (3.10)` and `python-test (3.13)`.

## AI disclosure

- [x] AI tooling was used. Tool and extent: Claude Code wrote the implementation and tests under my direction; I reviewed the diff.


pamod-madubashana (commented):
Bugfix review (audit fix):

Summary: `IsAvailable` now confirms KMS/vault actually holds the configured pubkey, not just metadata. Adds SPKI DER/PEM + raw parsers in `core`, wires AWS GetPublicKey, GCP PEM check, Fordefi vault address/compressed-key check. Correct identity-binding fix.

1. `PublicKeyFromSPKIDER/PEM` uses `x509.ParsePKIXPublicKey` + type-assert ed25519 - rejects RSA/ECDSA + truncated DER, good. Tests cover OID/truncation.
2. AWS `IsAvailable` now needs `kms:GetPublicKey` - permission change, document IAM update or availability will flap false.
3. Fordefi: `Address!=""` takes precedence over compressed key; empty/unknown shape returns false (fail-closed, good). Base64 std vs URL - confirm server uses std.
4. Extra KMS/vault round-trip on every `IsAvailable` - confirm caching/poll interval acceptable.

LGTM.

PR #315, opened 2026-09-15
feat: verify the configured public key belongs to the provider key

dev-jodee:
## What and why

The AWS KMS, GCP KMS, Turnkey, Vault and Fordefi backends all accept a Solana public key from config, never send it to the provider, and never check it against the key the provider actually holds. A mismatch surfaces only when the returned signature fails verification, after a wasted round trip; for `FordefiNativeAutoSigner`, after Fordefi has already broadcast.

Each backend's availability probe now confirms the configured key is the one the provider signs with.

| Backend | How |
|---|---|
| GCP KMS | `GetPublicKey` was already fetched and its `pem` discarded. Now compared. |
| Vault | `/v1/transit/keys/{name}` was already fetched and its `keys` discarded. Now compares the **latest** version, the one signing uses. |
| AWS KMS | New `GetPublicKey` call after `DescribeKey`. |
| Turnkey | New `get_private_key` query after `whoami`. Matches on the `ADDRESS_FORMAT_SOLANA` address or the hex public key. Skipped when `signWith` is already the Solana address. |
| Fordefi | The vault response was already fetched but only `id` was deserialized. Now reads the base58 `address` on chain-specific vaults, the base64 `public_key_compressed` on black box vaults. |

Applied identically in Rust, TypeScript, Python and Go. Shared Ed25519 SPKI decoding lands in each language's core (`signature_util` / `encoding.ts` / `core.sign`).

**Compatibility.** `IsAvailable` / `is_available` now calls `kms:GetPublicKey` on the AWS KMS path. Callers whose key policy does not grant that permission will see the health check start returning `false`. Signing is unaffected.

**Scope.** This is a health check, not a signing guard. Construction still makes no network calls anywhere and the trusted-provider model is unchanged. `is_available()` sits on no signing path, so this only helps callers who already run it. Every returned signature is still ed25519-verified against the configured key before it is attached.

Also removes a stale doc comment in `typescript/packages/fordefi/src/types.ts` claiming `FordefiVaultResponse` was used for "authoritative verification that a configured Solana `publicKey` actually belongs to the vault". No code implemented that until this change.

## Testing

- `just rust-test` — all three SDK matrices pass (395 / 395 / 400, 0 failed).
- `just rust-fmt` — `cargo fmt --check` and clippy clean.
- `just ts-build`, `just ts-test`, `just ts-fmt`, `just ts-treeshake` — all pass.
- Python was **not** run: no usable virtualenv on the dev machine (`ensurepip` unavailable). Files are syntax-checked only, and the `ruff format` diff CI reported has been applied. CI is the first real execution.
- Go was **not** run: no Go toolchain on the dev machine. Not compiled, not `gofmt`'d, not tested. CI is the first real execution.

## AI disclosure

- [x] AI tooling was used. Tool and extent: Claude Code wrote the implementation and tests under my direction; I reviewed the diff.


dev-jodee:
Superseded by #314. The public-key verification work is now a single commit on `fix/rust-audit` (6410e12) on top of the base-URL normalization commit, so both audit fixes review and merge as one PR. Closing this one; no code was dropped.

PR #317, opened 2026-09-17
ci: scope codeql analysis to production code

dev-jodee:
## What and why

This switches CodeQL from default setup to advanced setup, so that a config file with `paths-ignore` is actually honoured. Default setup ignores `.github/codeql/codeql-config.yml` entirely.

The config excludes test-only paths where test credentials and log statements recur and generate repeated non-findings: `rust/src/tests`, `rust/src/test_util.rs`, `python/tests` (plus the usual `**/node_modules` and `**/target`). All three source paths were verified to exist.

`rust/src/fireblocks/mod.rs` is deliberately left in scope. Its two dismissed `rust/cleartext-transmission` alerts are production custody-integration code, not test noise, and are worth a separate look rather than being filtered away.

Inline `#[cfg(test)]` modules under `rust/src/` cannot be path-filtered and remain scanned. That is intended: path filters work on files, and those test modules sit in the same files as production code.

Every language uses `build-mode: none`. For compiled languages, an autobuild/manual build mode causes `paths-ignore` to be silently ignored, so the filter would have no effect on the Rust and Go analyses.

Merging this alone is not sufficient: CodeQL default setup must be disabled in repo Settings > Code security. Otherwise both setups run and conflict.

No code scanning alerts were dismissed, closed, or modified by this change.

## Testing

Validated that both new YAML files parse:

```
python3 -c "import yaml,sys; [yaml.safe_load(open(f)) for f in sys.argv[1:]]" \
  .github/codeql/codeql-config.yml .github/workflows/codeql.yml
```

I also confirmed each `paths-ignore` source path exists in the tree, and that no other CodeQL workflow or config already exists in `.github/`.

I did not run CodeQL and I have not observed a CI run of this workflow. The first real validation will be the CodeQL run on this PR.

## AI disclosure

- [x] AI tooling was used. Tool and extent: Claude Code wrote the CodeQL config and workflow and drafted this description; I reviewed the diff.


dev-jodee:
will use org

PR #319, opened 2026-09-21
fix(cdp): send the network CDP needs to resolve address lookup tables

dev-jodee:
## Summary
- CDP's sign-transaction endpoint resolves address lookup tables against a specific network, so its optional `network` field is mandatory for a versioned transaction that references one. No language sent it, so such a transaction was rejected server-side with an opaque remote error.
- Adds an optional network setting to the CDP backend in Rust, TypeScript, Python and Go, validated against `solana` and `solana-devnet`, and included in the request body only when set.
- A transaction carrying lookup tables now fails locally with a config error when no network is configured, instead of issuing a request CDP will refuse.
- The default stays unset: mainnet versus devnet is not derivable from the transaction bytes, so a hardcoded default would be silently wrong for devnet callers.

## Test Plan
- `just rust-test` — 406 / 406 / 411 across the `sdk-v2`, `sdk-v3` and `sdk-v4` matrices.
- `pnpm --filter @solana/keychain-cdp test:unit` (34 passed), `typecheck` clean, `just ts-treeshake` clean.
- New per-language tests: unknown network rejected at construction, body carries `network` when set, body omits the key when unset, and a lookup-table transaction fails locally with no HTTP request issued.
- `just go-test` and `just py-test` were not runnable in my environment (no Go toolchain; Python venv could not be built). Both are unverified locally and rely on CI.

## Notes
- Stacked on #314 — review that one first; this PR targets `fix/rust-audit`.
- Rust needed a `v0` / `MessageAddressTableLookup` re-export in `sdk_adapter/v{2,3,4}.rs` to build a lookup-table test transaction.
- The TypeScript precheck fires on `version === 0` only: Kit's `V1CompiledTransactionMessage` has no `addressTableLookups` field, matching solana-go, which rejects lookups on v1 messages outright.


dev-jodee:
Folded these commits into the audit branch behind #314 to keep the audit fixes on a single branch. Superseded by the cherry-picked commits there.

